PR #54, opened 2026-06-12
feat(postage-usage): self-hosted batch utilization snapshots

### LINKED ISSUE #53: Self-hosted batch utilization snapshots (roaming issuer state)
## Problem

Issuing stamps from a postage batch requires per-bucket state: the next unused within-bucket index for each of the `2^u` collision buckets. Today this state is node-local. Losing it makes the batch unsafe to issue from, and there is no way to roam between machines with the same batch.

## Proposal

Store the utilization state inside the batch itself, as single-owner chunks at deterministic ids, stamped by the batch they describe:

- Snapshot chunk `n` has SOC id `keccak256("swarm-batch-usage" || batch_id || u16_be(n))`, owned by the batch owner. Chunk 0 is the root commit point; chunks `1..=L` are leaves holding the counter table.
- Because the ids never change, each snapshot chunk consumes exactly one storage slot forever: the reserve accepts a same-address, same-index re-stamp with a newer timestamp and replaces the SOC payload in place, regardless of batch mutability. The slots used by the snapshot chunks are recorded in the root, so the snapshot fully accounts for itself and the recursion (persisting the state changes the state) terminates at a bounded fixed point.
- Counters are encoded with patched frame-of-reference packing: a base, a bit width sized to the spread of the counters (not the batch depth), and an exception list (max 128) for outlier buckets. A half-full depth-24 batch fits in 15 chunks; an empty batch in a 78-byte root; the worst case is 65 chunks, under 0.05% of the smallest supported batch.
- Dilution changes no counter and no leaf byte, so nothing is reserved up front for growth that may never happen. The structure grows one leaf at a time only when the counter spread grows.
- The root carries a monotone sequence number and keccak digests of the leaf payloads, so readers always assemble a consistent snapshot or detect staleness. Counters are monotone, so elementwise max is a well-defined recovery merge for divergent copies; true multi-writer coordination is explicitly out of scope for v1.

Full format specification in the crate README.

## Scope

New crate `nectar-postage-usage`: `UsageTable` (counters, also a `StampIssuer` behind the `issuer` feature), `Snapshot::plan_persist` (self-accounting fixed point), two-phase decode (`RootInfo::parse` / `assemble`), and a `seal` feature turning plans into signed SOCs and stamps.

mfw78:
Closes #53

## What

New crate `nectar-postage-usage`: a compact, deterministic snapshot format for a batch's per-bucket slot counters, stored inside the batch itself as single-owner chunks at ids derived from the batch id and owner. A user can recover their issuer state on any machine from nothing but their key and batch id.

## How

Snapshot chunk `n` has SOC id `keccak256("swarm-batch-usage" || batch_id || u16_be(n))`. The ids never change, and the reserve accepts a same-address, same-stamp-index re-stamp with a newer timestamp (replacing the SOC payload in place, for mutable and immutable batches alike), so each snapshot chunk consumes exactly one storage slot for the lifetime of the batch. The root records the slots occupied by the snapshot chunks themselves; `Snapshot::plan_persist` runs this self-accounting to a bounded fixed point.

Counters are encoded with patched frame-of-reference packing: a base, a per-snapshot bit width sized to the counter spread (not the batch depth), and an exception list (max 128) so a single hot bucket cannot widen the whole table. Sizes: empty batch 78 bytes, half-full depth-24 batch 15 chunks, worst case 65 chunks (under 0.05% of the smallest supported batch). Dilution changes no leaf byte and reserves nothing ahead of time.

The root is the commit point: it carries a monotone sequence number and keccak digests of the leaf payloads, so a reader assembles a consistent snapshot or detects staleness. Decoding is strict (exact lengths, sorted exceptions, capacity checks, counter-sum check against the header total).

Format specification in `crates/postage-usage/README.md`.

## Surface

- `UsageTable`: counters plus geometry; `record`/`record_address`, `dilute`, `merge_max` (elementwise max as the divergence-recovery join). Implements `nectar_postage_issuer::StampIssuer` behind the `issuer` feature, so it drops into `BatchStamper`.
- `Snapshot::plan_persist` / `Snapshot::encode`, two-phase decode via `RootInfo::parse` / `RootInfo::assemble`.
- `usage_chunk_id` / `usage_chunk_address` deterministic addressing.
- `seal` feature: `seal_plan` turns a plan into signed `SingleOwnerChunk`s and `Stamp`s given the owner signer.

## Testing

22 tests: round-trips (inline, multi-leaf, exceptions, full-capacity), dilution byte-stability, steady-state slot reuse, self-accounting totals, corruption rejection (magic, truncation, digest, total, leaf count), issuer trait behavior, and seal verification (SOC address, owner recovery, stamp verify). `cargo fmt --all`, `clippy --all-targets --all-features -D warnings`, workspace check, and `--no-default-features` check all clean.


mfw78:
## Review: slot-collision and pre-mine analysis

Reviewed the "normal stamping cannot collide with snapshot slots by construction" claim against a threat model (can a content chunk take a metadata slot; can chunks be pre-mined to deny persist) and a balls-in-bins probability model. Bottom line: the design is sound for single-owner, frequent-persist, mutable-batch use, but one precondition is unstated and untested, and there is a real `BucketFull` availability tail at minimum depth.

### 1. The "by construction" invariant rests on a single-instance precondition the PR never wires up (high)

The collision-freedom proof holds only when **one `UsageTable` instance backs both content stamping and snapshot allocation**. `record` (`table.rs:157-170`) is the sole monotone allocator, and both the content path (`prepare_stamp -> record`, `issuer.rs:17-23`) and the snapshot path (`plan_persist -> table.record`, `snapshot.rs:178`) mutate the same `counts` Vec, so a recorded snapshot slot sits below the watermark forever.

But nothing wires a `Snapshot`'s table into a `BatchStamper` for content. `Snapshot` owns its table by value (`snapshot.rs:19`); `BatchStamper::new` consumes the issuer by value (`stamper.rs:113`); `MemoryIssuer`/`ShardedIssuer` are separate counters built from a `Batch`. An operator who stamps content via `MemoryIssuer::from_batch` while keeping a distinct `UsageTable` for snapshots has two independent counters and collision becomes possible. No test exercises the shared-instance topology.

Ask: add an adapter or documented constructor that lets one `UsageTable` back both a live `Snapshot` and content stamping, plus a test that stamps content and persists snapshots through the same instance and asserts no index overlap.

### 2. The real failure mode is `BucketFull` on lazy leaf allocation, worst at min depth (medium)

Leaf slots are allocated lazily the first time the widening spread needs them (`snapshot.rs:186-192`); the metadata address is deterministic and cannot be relocated. If that bucket is already at `2^(d-u)` when first needed, `record` returns `BucketFull` (`table.rs:163-164`) and persist aborts. nectar hard-errors and never wraps to 0, unlike bee's mutable-batch wrap (`stampissuer.go:186-193`), so on an immutable batch the wedge is permanent; on a mutable batch recovery is operator-driven cross-address re-stamping.

Probability of a blocked persist, `f = P(Poisson(rho*C) >= C)`, `P_block = 1-(1-f)^K`:

- **Deferred first-persist, d=17 (C=2):** P_block at K=1 is 0.26 / 0.54 / 0.59 at rho 0.50 / 0.90 / 0.99. The only genuinely dangerous regime.
- **Collapses with depth:** d=24 K=65 at rho=0.8 is 2e-2; d=29 is ~0.
- **Realistic frequent-persist: negligible everywhere** (root P_block ~2e-8, leaves <=~2e-13) because leaves are allocated while their buckets are near-empty.

(Closed form and seeded Monte Carlo agreed within sampling noise.)

Ask: persist early on d<=20, and/or reserve the root slot at batch creation (drops root P_block to ~2e-8). Optionally offer an opt-in `allocate_n_leaves` / pre-persist `has_capacity` sweep for deployments that cannot guarantee frequent persist.

### 3. Pre-mine is externally impossible, insider-only

A valid stamp requires `recovered_signer == owner` (`stamp.rs:315-324`) and the wire stamp has no delegation field, so a non-owner cannot stamp onto the batch; cross-owner DHT collision is ~2^256. The only pre-mine vector is a key holder (shared/delegated/gateway batch): metadata buckets are publicly computable, grinding a content chunk into a target 16-bit bucket costs ~2^16 keccak trials, and filling it costs `2^(d-16)` chunks (trivial at low depth). Worth a note in the spec that the snapshot inherits the batch's single-owner trust boundary and is not safe to expose to untrusted stampers on a shared key.

### 4. `reserved_stamp_indices` / `is_reserved` are advisory only

Consulted by no allocation, stamping, or eviction path (`snapshot.rs:125-142`). Fresh issuance is auto-protected, but a caller doing cross-address re-stamping on a mutable batch who forgets `is_reserved` evicts a snapshot chunk and corrupts persisted state. Either route re-stamping through a method that checks reservation automatically, or document the requirement at the call site. Minor: `is_reserved` rebuilds the full list per call (O(n)).

### Nits

- `plan_persist` allocation is correct and atomic: `*self = work` only on success (`snapshot.rs:216`), fixed point breaks at `slots.len() > leaves.len()`, no off-by-one. No issue.
- `dilute` / `merge_max` preserve monotonicity (only raise counters / depth). No issue.
- Document the nectar-vs-bee `record` divergence (hard-error vs wrap-to-0); operators relying on bee behaviour will be surprised.

Net residual risk: low. Items 1 and 2 are the ones worth acting on before merge.

PR #55, opened 2026-06-13
feat(primitives): gate wasm-bindgen-rayon behind default-off `wasm-threads` feature

v1rtl:
Currently `nectar-primitives` pulls `wasm-bindgen-rayon` unconditionally on `wasm32` and re-exports `init_thread_pool` from `src/wasm.rs`. The **mere presence** of that re-export forces wasm-bindgen's threads transform, which requires the module to use a **shared** (`SharedArrayBuffer`) memory. In a browser that means the hosting page must be **cross-origin isolated** (`COOP: same-origin` + `COEP: require-corp`).

That isolation requirement is a hard blocker for hosts that cannot set those response headers — notably **ENS gateways like eth.limo / bzz.limo** (they only send `Cross-Origin-Resource-Policy`, never COOP/COEP), and many static hosts. There is currently no way to build a `nectar-primitives` wasm that runs without shared memory.

### Change

Gate the thread pool behind a new **default-off** `wasm-threads` feature:

- `wasm-bindgen-rayon` becomes `optional = true`.
- `wasm-threads = ["dep:wasm-bindgen-rayon"]`.
- The `init_thread_pool` re-export in `src/wasm.rs` is `#[cfg(feature = "wasm-threads")]`.

Default builds are unchanged in API; they simply no longer pull `wasm-bindgen-rayon`, so the resulting wasm uses a plain (non-shared) linear memory and needs no cross-origin isolation. Consumers that want threaded BMT hashing opt in with `--features wasm-threads` (plus the usual `+atomics`/`--shared-memory` rustflags).

It is **default-off on purpose**: dependent crates (`nectar-postage`, `nectar-mantaray`, `nectar-postage-issuer`) depend on `nectar-primitives` with default features, so a default-on threads feature would be forced on transitively and could not be opted out by a downstream binary.

The plain `rayon` usage elsewhere (`bmt`, `file/sync_splitter_parallel`, `file/sync_joiner`) is untouched — on wasm without a thread pool it runs inline, which is correct (just single-threaded).

### Verification

- `cargo check -p nectar-primitives --target wasm32-unknown-unknown` (default): compiles, no `wasm-bindgen-rayon` in the graph, produces a non-shared memory.
- `--features wasm-threads`: pulls `wasm-bindgen-rayon` as before (threaded/shared-memory build).
- Verified end-to-end in a downstream wasm consumer: default build links without `--shared-memory`/`+atomics` and runs single-threaded; `--features wasm-threads` build imports a shared memory as before.

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -65,7 +65,12 @@ alloy-primitives = { workspace = true, features = ["asm-keccak", "getrandom"] }
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 alloy-primitives = { workspace = true, features = ["getrandom"] }
 wasm-bindgen.workspace = true
-wasm-bindgen-rayon = "1.3"
+# Gated behind the `wasm-threads` feature (default-off). Its presence forces
+# wasm-bindgen's threads transform, which requires a SharedArrayBuffer memory
+# (and thus COOP/COEP cross-origin isolation on the hosting page). Omitting it
+# yields a wasm that needs no shared memory; the plain `rayon` usage elsewhere
+# then runs inline (no thread pool) on wasm.
+wasm-bindgen-rayon = { version = "1.3", optional = true }
 js-sys.workspace = true
 getrandom.workspace = true
 getrandom_02.workspace = true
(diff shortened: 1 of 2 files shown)


mfw78 (changes requested):
Looks straight-forward and good. Just a minor request to include a feature of 'perf' or something like this (no firm opinion on terminology), and that this feature would then bring in wasm-thread). Just a DX issue to make sure consumers of the library understand with more plain terminology what's happening / how to use it.

Also, can you please follow the CLA bot to sign the CLA. Cheers.

v1rtl:
I have read the CLA Document and I hereby sign the CLA

v1rtl:
> Looks straight-forward and good. Just a minor request to include a feature of 'perf' or something like this (no firm opinion on terminology), and that this feature would then bring in wasm-thread). Just a DX issue to make sure consumers of the library understand with more plain terminology what's happening / how to use it.

I don't know what you really mean. It's not about perf but enabling conditional compilation

mfw78:
Thanks for this, @v1rtl, and for the discussion on naming. Your commit has landed in #80, cherry-picked with your authorship preserved and with you as co-author on the follow-up. There I also added a `parallel` feature that turns on `wasm-threads`, matching the `parallel` feature already used by `nectar-postage` and `nectar-postage-issuer` so the terminology is consistent across the workspace. Closing this in favour of #80. Really appreciate the contribution, and the cross-origin-isolation insight about eth.limo that motivated it.

Issue #70, opened 2026-06-14
Network-validation wave: persist-time sequence CAS to complete T3 from #56


mfw78:
The network-validation wave: enforce snapshot sequence monotonicity at persist time, completing the protocol-level half of T3 from #56.

## Context

#56 (merged via #68) made the in-memory downgrade unrepresentable: no public API yields an owned `UsageTable` from a recovered snapshot, so a recovered snapshot cannot be rebuilt at sequence 0 and re-persisted in memory. #65 (merged via #67) delivered the `SnapshotStore` trait, the local cache seam.

What neither closes, by design, is the protocol-level downgrade: the reserve overwrites a chunk by stamp timestamp, not by snapshot sequence, so a persist that stamps the metadata chunk with a newer timestamp but a lower or reset sequence overwrites a newer published snapshot in place. The only authority on the currently published sequence for a batch is the network (the live root chunk), or a trusted monotonic store. The #56 rustdoc and README disclose this honestly and point here.

## Requirement

Before publishing a persist, read the currently published root chunk for the batch (via `SnapshotStore` as a cache plus a network read) and reject the persist unless its sequence strictly exceeds the published sequence: a compare-and-swap on the sequence keyed at the metadata chunk address. The store load path should expose the published sequence, or a `Validated` wrapper carrying a sequence floor, so the issuer enforces this without a separate fetch. This is the validation seam the #56 design deliberately left a hook for.

Also in scope, as the same wave wires the issuer to the network and the store:
- (a) Fresh-construction forgery: a fresh table minted from the public constructors and persisted at sequence 0 for a batch that already has a published higher sequence is caught by the same compare-and-swap.

## Relationship to the other issuer work

This is downstream of #62 (mutable ring issuance) and #63 (unify the issuer counter table). The compare-and-swap lands once the issuer consumes `SnapshotStore` and the network read path. Tracks the T3 residual carried out of #56 and #65.

mfw78:
Delivered via #74 (merged into the foundation): a Snapshot validation type-state. A fresh or cache-loaded snapshot is unvalidated and cannot persist; `Snapshot::revalidate(floor: PublishedSequence) -> Validated` is the only route to a `PersistPlan` and enforces next sequence > floor, closing both the fresh-construction forgery and the stale recovered/cache persist. The published floor is a live network read of the root chunk's sequence, supplied by the consumer (vertex); nectar provides the enforcement. This completes the protocol half of T3 that #56 deferred.

PR #78, opened 2026-06-15
feat(postage-usage): BatchStamper open/stamp/flush client facade

### LINKED ISSUE #77: [postage-usage] High-level BatchStamper facade over a ChunkSource/ChunkSink retrieval seam
The cross-machine roam in `nectar-postage-usage` is correct but ceremonious: the example threads ~16 calls across ~11 types, hand-managing the owner, batch id, every chunk address (with a `+1` leaf offset), the published-sequence floor, `RootInfo::parse`/`assemble`, the seal timestamps, and the upload. Issuing a stamp without losing state should be three calls, not sixteen.

## Proposal

Add a high-level, async facade behind a new `client` feature, purely additive over the existing low-level API (which stays fully public for power users):

```rust
let mut stamper = BatchStamper::open(signer, &batch, source, sink).await?; // recover or start fresh
let idx = stamper.stamp(&content_address)?;                                // issue
stamper.flush().await?;                                                     // revalidate + seal + upload
```

The network access goes through two small traits defined in this crate (so vertex, the consumer, implements them; nectar never depends on vertex, no dependency cycle):

- `ChunkSource::fetch(&address) -> Result<Option<Bytes>, E>` reads a chunk's payload from the network. `Ok(None)` means the chunk is definitively absent; `Err` means the read could not be completed. This distinction is load-bearing: treating a transport failure as absence would read the published-sequence floor as `NONE` and reopen the downgrade #70 closes.
- `ChunkSink::push(&sealed) -> Result<(), E>` publishes a sealed snapshot chunk.

Both mirror the async style of `nectar_primitives::ChunkGet` (so they are wasm-compatible and `auto_impl`-derivable), and both are trivially implementable by vertex's `SwarmChunkProvider` (`retrieve_chunk(&ChunkAddress)`), which already takes a `SwarmAddress` and returns a chunk wrapping `nectar_primitives::AnyChunk`.

The facade owns the floor read, the leaf loop, the seal clock, and the `Snapshot::new`-vs-`from_parts` foot-gun. `Snapshot`, `RootInfo`, `revalidate`, `PublishedSequence`, `seal_plan`, and the address derivations stay public.

## Follow-up (separate, vertex repo)

Implement `ChunkSource`/`ChunkSink` for vertex's chunk provider once this lands and the nectar rev is bumped. Out of scope for this nectar PR.

mfw78:
## What

Adds a high-level async facade, `BatchStamper`, to `nectar-postage-usage` behind a new, purely additive `client` feature. The facade collapses the multi-step, cross-machine roam (parse the published root, fetch every leaf at its `+1` offset, assemble a snapshot, revalidate against the live floor, plan a persist, seal each chunk, push each chunk: roughly sixteen low-level calls) into three: `open`, `stamp`, `flush`.

- `BatchStamper::open(signer, batch, source, sink)` binds the signer, owner, and batch id once, recovers a published snapshot from the network, or starts fresh only when the root is definitively absent.
- `BatchStamper::stamp(content)` is synchronous and local: no network, no `.await`.
- `BatchStamper::flush()` re-reads the live floor, revalidates, plans, seals with a strictly increasing timestamp, and pushes every sealed chunk.

Two new async traits, `ChunkSource` and `ChunkSink`, supply the network. They are defined in this crate so that a node (vertex) implements them; nectar does not depend on the node. Both use the RPITIT `impl Future<...> + Send` style to match the primitives `ChunkGet` convention and stay wasm-friendly, and both carry `#[auto_impl::auto_impl(&, Arc, Box)]`. `ClientError` is a `thiserror` enum generic over the source and sink error types, unifying `UsageError` and `SealError`.

The low-level `Snapshot` / `revalidate` / `plan_persist` / `seal_plan` / `usage_chunk_address` API is unchanged and remains public for power users. The example was rewritten over the facade with an in-memory `MemNet` implementing both traits.

## Why

Closes #77.

## Testing

- The `roam_between_machines` example runs to completion, printing a legible machine A to machine B roam over a shared `MemNet`.
- Floor-safety tests: `open_aborts_on_source_error` and `flush_aborts_on_floor_read_error` prove a transport `Err` aborts rather than degrading into a fresh start or a `PublishedSequence::NONE` floor; the latter also asserts nothing is published.
- Lifecycle tests: `open_miss_starts_fresh_and_flush_publishes`, `open_recovers_published_batch`, `stamp_then_flush_advances_sequence_and_reuses_slots`, and `flush_rejects_stale_sequence`.
- SBU1 golden vectors (`readme_worked_example_vector`, `readme_large_batch_multi_leaf_vector`, `mutable_vector_flags_byte_and_round_trip`) are byte-untouched and green; `codec.rs`, `snapshot.rs`, `seal.rs`, and `table.rs` are not modified.
- `cargo fmt`, `cargo clippy --all-targets --all-features -D warnings`, `cargo test --all-features`, and `cargo rustdoc -D warnings` are all clean. Default and seal-only (no client) builds also verified clean, confirming additivity.

## Security

The `ChunkSource::fetch` contract is load-bearing and enforced: `Ok(None)` means definitively absent, and `Err` means the read did not complete. In `open`, the root and leaf reads propagate `Err` via `.map_err(ClientError::Source)?` before any `Snapshot::from_batch` fresh-start can run, so only literal `Ok(None)` starts fresh; a missing leaf for a published root is `ClientError::MissingLeaf` (corruption), never a fresh start. In `flush`, the live floor re-read propagates `Err` before revalidate, seal, or push run, so a failed read never persists; only `Ok(None)` maps to `PublishedSequence::NONE`. This preserves the #70 downgrade protection. `revalidate` enforces that the next sequence strictly exceeds the floor.

There is no nectar to vertex dependency and no cycle: the new traits add no edge back toward the node, `auto_impl` is an optional crate dependency gated behind `client`, and `tokio` is a dev-dependency only.

## Follow-up

The vertex-side `ChunkSource` / `ChunkSink` implementation (over retrieval and pushsync) is out of scope here and pending the nectar rev bump in the node workspace.

## AI Assistance

Produced by an automated multi-agent workflow and human-reviewed before merge.

### DIFF
--- a/crates/postage-usage/src/client.rs
+++ b/crates/postage-usage/src/client.rs
@@ -27,7 +27,8 @@
 //! a batch that already has a published root.
 
 use alloc::vec::Vec;
-use std::time::{SystemTime, UNIX_EPOCH};
+
+use web_time::{SystemTime, UNIX_EPOCH};
 
 use alloy_primitives::Address;
 use alloy_signer::SignerSync;
@@ -60,10 +61,15 @@ pub trait ChunkSource {
 
     /// Fetches the data payload of the chunk at `address`, or `Ok(None)` if the
     /// network confirms no such chunk exists.
+    ///
+    /// The returned future intentionally carries no `Send` bound: a native
+    /// transport whose future is `Send` propagates that through
+    /// [`BatchStamper::open`] and [`BatchStamper::flush`] automatically, while a
+    /// browser transport with a `!Send` future can still implement this trait.
     fn fetch(
         &self,
         address: &SwarmAddress,
-    ) -> impl core::future::Future<Output = Result<Option<Bytes>, Self::Error>> + Send;
+    ) -> impl core::future::Future<Output = Result<Option<Bytes>, Self::Error>>;
 }
 
 /// Publishes a sealed snapshot chunk (the single-owner chunk and its stamp) to
@@ -74,10 +80,14 @@ pub trait ChunkSink {
     type Error: core::error::Error + Send + Sync + 'static;
 
     /// Uploads a sealed snapshot chunk.
+    ///
+    /// As with [`ChunkSource::fetch`], the returned future carries no `Send`
+    /// bound so a `!Send` browser transport can implement it; native `Send`-ness
+    /// propagates through the facade on its own.
     fn push(
         &self,
         sealed: &SealedChunk,
-    ) -> impl core::future::Future<Output = Result<(), Self::Error>> + Send;
+    ) -> impl core::future::Future<Output = Result<(), Self::Error>>;
 }
 
 /// Errors produced by the [`BatchStamper`] facade.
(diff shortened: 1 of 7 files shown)


crates/postage-usage/src/client.rs
```diff
@@ -0,0 +1,567 @@
...
+//! addresses. That guarantee survives only if a failed network read is never
+//! mistaken for "the chunk is absent". [`ChunkSource::fetch`] therefore
+//! distinguishes `Ok(None)` (the network definitively agrees the chunk does not
+//! exist) from `Err` (the read could not be completed). [`BatchStamper::open`]
+//! and [`BatchStamper::flush`] both abort on `Err`: a transport failure never
+//! becomes a fresh sequence-0 start nor a [`PublishedSequence::NONE`] floor over
+//! a batch that already has a published root.
+
+use alloc::vec::Vec;
+use std::time::{SystemTime, UNIX_EPOCH};
```
mfw78: This is a massive footgun for WASM compatibility. Suggest perhaps moving wasm compatibility layer handlers for time from vertex to nectar to ensure the primitives can be built safely for both targets.

Issue #92, opened 2026-06-15
feat(feeds): support reading legacy v1 feed payloads (timestamp + reference)


mfw78:
Deferred to Post v1 under epic #693. Split out of the initial `nectar-feeds` work: #83 is closed and landed the crate, and #84 carries the epoch indexing. The layer-2 chunk primitives epic #82 that this issue was filed under is closed.

## Context

Swarm feed update payloads come in two encodings. The encodings are orthogonal to the indexing scheme, are detected by length, and are resolved by a dual-resolution race on read.

- v1 (legacy): the post-span payload is timestamp(8) plus reference(32), optionally plus a 32-byte encryption key, so 48 or 80 bytes in total. The feed update points at another chunk and carries a publish timestamp.
- v2 (current): bare arbitrary data. This is what modern writers emit, and it is the only encoding `nectar-feeds` models.

The shipped crate is v2 only: the SOC payload is opaque bytes. For full read interop with older feeds that are still on the live network, a client must also decode v1 payloads.

## Scope

- A typed payload model, for example `FeedPayload::{ V1 { timestamp: u64, reference: SwarmAddress, encryption_key: Option<...> }, V2(Bytes) }`, with length-based detection: 48 or 80 bytes is v1, anything else is v2.
- The reader decode path returns the typed payload. The writer continues to emit v2 only, because a new client should not write a dead format.
- The dual-resolution race, and the resolution of a v1 reference to actual content, stay in the vertex client layer because they need retrieval. They do not belong in `nectar-feeds`. This issue is the pure payload decoder only.
- Conformance vectors for both v1 lengths.

## Decision record

v1 write support is out of scope on purpose, because writing a legacy format is not something a new client should do. MRU, the pre-Swarm mutable resource updates, is unsupported.

mfw78:
Still valid and correctly scoped as a follow-on to #83/#84 rather than a duplicate: #83/#84 are the indexing schemes, this is the payload encoding, and the two are orthogonal. Nothing has landed. Note that #82 sequencing does not list this issue.

Issue #82, opened 2026-06-15
epic: layer-2 chunk primitives over CAC/SOC (feeds, PSS, ACT)


mfw78:
## Problem

Swarm's higher-order data primitives (feeds, PSS messaging, GSOC, and the access-control trie) are all compositions of the two wire chunk types, content-addressed chunks (CAC) and single-owner chunks (SOC), plus a small amount of cryptography. None of them is a new wire-level chunk type. On the wire they remain CAC (type 0) or SOC (type 1); the "feed-ness" or "trojan-ness" is a structured interpretation of an existing chunk, modelled as builder plus interpreter views.

These belong in nectar because none requires a network or a database. They mirror the existing `nectar-mantaray` and `nectar-postage` crates and must be usable by any Swarm consumer (indexer, web tool, light client) without pulling in a libp2p stack.

## End state

Three new crates — `nectar-feeds`, `nectar-pss`, `nectar-accesscontrol` — plus one ECIES module in `nectar-primitives`, each wasm-friendly, network-free, and byte-conformant with the reference network.

## Substrate that already exists

`nectar-primitives` already provides the foundation:

- `ContentChunk` (CAC, `address = bmt_hash(span || payload)`) and `SingleOwnerChunk` (SOC, `id` + 65-byte signature over `keccak(id || body.hash())`, owner recovered from signature, `address = keccak(id || owner)`).
- `BmtBody<BODY_SIZE>` where `body.hash()` equals the wrapped CAC address. This means SOC already signs the same digest the reference network signs, so feed and envelope SOCs are byte-conformant with no change to SOC code.
- Keccak-256 CTR symmetric cipher (`chunk/encryption`), the `Getter`/`Putter` store traits, and dispersed-replica SOC support.

## Approach

1. No new `ChunkTypeId`. Each primitive is a constructor over `SingleOwnerChunk`/`ContentChunk` plus an interpreter that validates and reads one back.
2. One crate per family, matching `nectar-mantaray`/`nectar-postage`.
3. Carry the `BODY_SIZE` const generic through every new chunk-producing type.
4. Finders and ACT are generic over the `store` `Getter`/`Putter` so they stay wasm-friendly and testable in memory.

[...]

PR #93, opened 2026-06-15
feat(feeds): Swarm feeds (sequence + epoch) over single-owner chunks

### LINKED ISSUE #39: [FEATURE] WASM-friendly ChunkGet variant without Send bound
## What problem does this solve?
`nectar-primitives::store::ChunkGet` requires `+ Send` on its returned futures (`crates/primitives/src/store/typed.rs:79-103`). On `wasm32-unknown-unknown`, the underlying primitives are not `Send`: `tokio_with_wasm::time::{Sleep, Timeout, Interval}` are `!Send`, and libp2p swarm + stream futures are `!Send` in single-threaded targets.

Today every WASM consumer wraps each `fetch_chunk` call with `send_wrapper::SendWrapper` to coerce the type-system; isheika does this at `client.rs:696-712` with a load-bearing safety comment. The wrapper compiles because WASM is single-threaded so the `Send` lie is unobservable, but it is a static-type hack that should not be required to satisfy the trait.

Same friction surfaces in `FuturesUnordered` callers, which must cfg-gate `BoxFuture` (native, `Send`) vs `LocalBoxFuture` (WASM, `!Send`); isheika does this at `transport.rs:1933-1941`.

## Describe the solution
Either:

(a) Add a `LocalChunkGet` trait with the same method shape but no `Send` bound, plus a blanket `impl LocalChunkGet for T: ChunkGet` so existing call sites keep working. Mantaray traversal etc. can then bound on `LocalChunkGet` and accept both.

(b) Cfg-gate the `Send` bound off on `wasm32`. Less surface area but bleeds the cfg into trait callers.

(a) is cleaner: it lets a single Rust crate ship code paths that work on both native (over a `Send` future executor) and WASM (over a `!Send` browser executor) without per-target `impl` blocks.

## Alternatives considered
- `send_wrapper::SendWrapper` everywhere (today's reality): unsafe to extend beyond WASM-single-threaded, and pollutes every downstream.
- Per-target `impl` blocks gated by `cfg(target_arch = "wasm32")`: works but doubles the surface area in every consumer (see isheika's split at `client.rs:2946-2981`).
- A `LocalChunkGet` newtype wrapper at the consumer end: same hack, just relocated.

## Implementation notes
- The other trait families in `crates/primitives/src/store/` (`ChunkPut`, `ChunkHas`, ...) likely need the same treatment for consistency; tackle them together.
- `Chunk` itself is `Send + Sync + 'static` (`crates/primitives/src/chunk/traits.rs:48`). That may need the same conditional relaxation depending on what the WASM target requires; out of scope for this issue but related.
- Once `LocalChunkGet` exists, mantaray's async traversal (see #37) can bound on it directly, and WASM mantaray traversal works without the `BlockingNetworkedStore` adapter.

## Are you willing to PR this?
Maybe later. Probably wants paired design with #37 since they share the WASM motivation.

AI Assistance: Claude Code used for analysis and authoring this issue.

### LINKED ISSUE #83: feat(feeds): nectar-feeds crate (Feed, Index, sequence indexing, updater, getter)
Part of the layer-2 chunk primitives epic.

## Goal

Scaffold a new `nectar-feeds` crate and implement the feed data model plus sequence (numeric) indexing end to end: construct, sign, address, read, and update a feed.

## Composition

A feed is identified by a `(topic, owner)` pair. Each update is a SOC whose body wraps a CAC of the payload, and whose `id` is derived from the topic and an index:

```
id      = keccak(topic || index.marshal())
address = keccak(id || owner)            // standard SOC address
```

Because `nectar-primitives` SOC already signs `keccak(id || body.hash())` and `body.hash()` is the wrapped CAC address, no change to SOC code is required.

## Wire-critical detail

Sequence index marshals to `start_be_u64`, 8 bytes, NOT hashed (contrast with epoch indices, which are hashed; see the epoch issue). This asymmetry matches the reference implementation and must be preserved.

## Scope

- [ ] New crate `crates/feeds` (`nectar-feeds`), wired into the workspace, building on `wasm32-unknown-unknown`.
- [ ] `Index` trait (`marshal`, `next`), `Sequence(u64)` implementation.
- [ ] `Feed { topic, owner }` with `update_id` and `update_address`, carrying `const BODY_SIZE: usize = DEFAULT_BODY_SIZE`.
- [ ] `Getter` (read an update by index over `store::Getter`) and `Putter`/updater (wrap payload as CAC, wrap as SOC, sign, write over `store::Putter`).
- [ ] Sequence finder: synchronous linear scan returning the latest present chunk and next index; async exponential-then-binary search variant.
- [ ] Conformance vectors under `tests/`: pin address and signed bytes for a fixed topic and owner key at index 0 and a non-zero index.

## Dependencies

Foundational. The epoch-indexing issue builds on this crate scaffold.

Part of #82.

### LINKED ISSUE #84: feat(feeds): epoch-grid indexing and sync/async finders
Deferred to Post v1 under epic #693. The layer-2 chunk primitives epic #82 that this issue was filed under is closed.

## Goal

Add epoch-grid indexing and its finders to `nectar-feeds`, so a lookup of sporadic (unpredictable) updates does not need an exhaustive scan.

## Composition

An epoch is a `(start, level)` pair that covers `[start * 2^level, (start + 1) * 2^level)`. Level ranges 0 to 32, which is 1-second granularity up to roughly 136 years. The base time at a level is the timestamp with its `level` least-significant bits cleared. An update is written into the longest free epoch that contains its timestamp.

## Wire-critical detail

Epoch index marshals to `keccak(start_be_u64 || level_u8)`, a 32-byte value, which then feeds the feed `id = keccak(topic || index)`. Note the contrast with sequence indexing, which is raw big-endian and unhashed.

## Correction: what the crate now looks like

The scaffold this issue waited for has landed. #83 is closed and `crates/feeds` ships `Feed`, the `Index` trait, `Sequence`, and the read and write handles. Two facts in the original text no longer describe the tree.

The handles are `Reader` (`crates/feeds/src/reader.rs`) and `Publisher` (`crates/feeds/src/publisher.rs`), not a getter and a putter. The store seam they run over is `ChunkGet` in `nectar_primitives::store`, and there is no `store::Getter`.

The shipped `Index` trait (`crates/feeds/src/index.rs`) is `marshal()` plus `next()`, and `next()` takes no arguments. Epoch selection needs `(last, at)`, so it does not fit the trait as it stands. The trait doc already records this: it says the seam stays open for a future epoch index whose lookup is a different search and not a drop-in for the sequence finders. Decide whether to widen `Index` or to add a second seam before any code is written.

## Scope

- [ ] `Epoch { start, level }` with the epoch-selection logic, which picks the longest free epoch that contains the timestamp. Decide first how it binds to the `Index` seam, per the correction above.
- [ ] Synchronous epoch finder: walk from the lowest common ancestor of `(at, after)` up the grid until a stored chunk is found, then descend to the finest-resolution match.
- [ ] Asynchronous epoch finder: probe epochs concurrently and recursively narrow the bracket around the target time. It must stay runtime-agnostic, with no tokio dependency, so it compiles to wasm over the `ChunkGet` future. Concurrent probing runs on `FuturesUnordered` plus the governor, which is the one sanctioned fan-out substrate.
- [ ] Epoch updater that advances `(last, epoch)` after each put.
- [ ] Conformance vectors under `tests/`: pin the address and the signed bytes for an epoch update at a fixed `(start, level)`, topic, and owner key.

mfw78:
## Summary

Adds `nectar-feeds`, implementing Swarm feeds as single-owner chunks, together with the nectar-primitives support it needs. A feed update is a single-owner chunk with id `keccak256(topic || index)` at address `keccak256(id || owner)`; the chunk body wraps a content-addressed chunk of the payload.

## What is included

`feat(primitives)`: relax the async store-trait Send bound to a target-conditional `MaybeSend`/`MaybeSync` so wasm32 consumers no longer need a `SendWrapper` hack while native keeps the Send guarantee; add `ChunkGet::is_not_found` so a genuine miss is distinguishable from a real backend error; make nectar-primitives build for `wasm32-unknown-unknown`.

`feat(feeds)`: the crate itself. Sequence indexing is the canonical scheme; epoch indexing is provided behind the `epoch` feature as experimental. The reader verifies the single-owner chunk address and binds the recovered owner to the feed owner. The writer signs the chunk and advances a scheme-specific write cursor that selects the publish index from the update time, so the time-indexed epoch scheme is driven correctly through the generic writer. Latest-update lookup is a per-scheme async finder over `ChunkGet` with no wall clock in the crate; the caller supplies the target time.

## Versioning decisions

Sequence is the canonical scheme. Epoch is experimental behind its feature (it is not used on the live network). Payloads are v2 bare-data only; reading the v1 legacy payload (timestamp plus reference) is deferred to #92. MRU is unsupported.

## Verification

`cargo fmt` and `cargo clippy --all-targets --all-features -- -D warnings` are clean. 32 feeds tests and the full primitives suite pass. nectar-feeds compiles on `wasm32-unknown-unknown` (default and `epoch`). Zero unsafe in the crate. Conformance vectors pin the single-owner chunk anchor, the topic and id derivation, and a reference-captured full signed-chunk fixture.

## Review

Designed across independent architecture reviews and hardened through a multi-agent audit (code quality, wire conformance, security, attack and QA) with adversarial verification of every finding. A notable fix: the epoch read path now runs the same address verification as the sequence path, so a foreign or wrong-epoch owner-signed chunk returned by an untrusted store is rejected rather than accepted as the queried slot.

Closes #39
Closes #83
Closes #84
Part of #82
Follow-up: #92

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2060,6 +2060,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "nectar-feeds"
+version = "0.2.1"
+dependencies = [
+ "alloy-primitives",
+ "alloy-signer",
+ "alloy-signer-local",
+ "arbitrary",
+ "bytes",
+ "derive_more",
+ "futures",
+ "nectar-primitives",
+ "proptest",
+ "serde",
+ "strum 0.28.0",
+ "thiserror",
+ "tokio",
+]
+
+[[package]]
 name = "nectar-mantaray"
 version = "0.2.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ nectar-swarms = { version = "0.2.1", path = "crates/swarms" }
 nectar-postage = { version = "0.2.1", path = "crates/postage" }
 nectar-postage-issuer = { version = "0.2.1", path = "crates/postage-issuer" }
 nectar-postage-usage = { version = "0.2.1", path = "crates/postage-usage" }
+nectar-feeds = { version = "0.2.1", path = "crates/feeds" }
 
 # alloy
 alloy-primitives = { version = "1.6", default-features = false }
--- a/crates/feeds/Cargo.toml
+++ b/crates/feeds/Cargo.toml
@@ -1,0 +1,58 @@
+[package]
+name = "nectar-feeds"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+exclude.workspace = true
+description = "Swarm feeds over single-owner chunks for Ethereum Swarm"
+documentation = "https://docs.rs/nectar-feeds"
+readme = "README.md"
+keywords = ["swarm", "ethereum", "feeds", "soc"]
+categories = ["data-structures", "cryptography::cryptocurrencies"]
+
+[lints]
+workspace = true
+
+[dependencies]
+nectar-primitives = { workspace = true }
+alloy-primitives = { workspace = true }
+alloy-signer = { workspace = true }
+bytes = { workspace = true }
+thiserror = { workspace = true }
+strum = { workspace = true }
+derive_more = { workspace = true }
+futures = { workspace = true }
+
+# optional
+serde = { workspace = true, optional = true }
+arbitrary = { workspace = true, features = ["derive"], optional = true }
+
+[dev-dependencies]
+proptest = { workspace = true }
+alloy-signer-local = { workspace = true }
+alloy-primitives = { workspace = true, features = ["getrandom"] }
+tokio = { workspace = true, features = ["rt", "macros"] }
+
+[features]
+default = ["std"]
+
+# Standard library support. The async reader/writer paths build on the typed
+# chunk store traits, which require `std`.
+std = ["nectar-primitives/std", "alloy-primitives/std", "thiserror/std", "serde?/std"]
+
+# Serialization support with serde.
+serde = ["dep:serde", "nectar-primitives/serde", "alloy-primitives/serde"]
+
+# Arbitrary trait implementations for property-based testing.
+arbitrary = ["dep:arbitrary", "nectar-primitives/arbitrary", "alloy-primitives/arbitrary"]
+
+# Epoch-based feed scheme (time-indexed lookup over a binary epoch grid).
+epoch = []
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/crates/feeds/src/epoch.rs
+++ b/crates/feeds/src/epoch.rs
@@ -1,0 +1,934 @@
+//! Epoch feed scheme: a time-indexed binary grid.
+//!
+//! An epoch is a half-open time interval `[start, start + 2^level)` on a binary
+//! grid. The scheme lets a reader find the update live at any timestamp with a
+//! logarithmic walk, at the cost of a stateful write cursor that must remember
+//! the previous update time to place the next epoch.
+//!
+//! # Grid math
+//!
+//! Each epoch spans `2^level` seconds. An epoch is partitioned into a left and
+//! a right child one level down; the child whose interval contains a target
+//! timestamp is selected by testing the `2^(level-1)` bit of the timestamp. The
+//! root epoch `{start: 0, level: 32}` covers the whole representable grid.
+//!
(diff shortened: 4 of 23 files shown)


mfw78:
Superseded by #479 (merged), which implements the sequence-indexing feeds crate fresh against the current chunk API, the nectar-testing harness, and the current arbitrary/proptest conventions rather than this branch's June-era idioms. This branch was read for salvageable structure and test vectors during that work.

Epoch indexing (the other half of this PR's title) was deliberately out of scope for #479 and continues as issue #84; the SOC-typed getter and windowed-finder follow-ups are #511/#512. Closing as superseded; nothing here is lost.

PR #97, opened 2026-06-17
feat(primitives): add redistribution sampling + entitlement primitives

mfw78:
## Summary

Adds a `nectar_primitives::redistribution` module providing the pure-compute primitives that underpin the storage-incentives redistribution game. These are the building blocks for vertex's storer `RedistributionStore`, namely uniform neighbourhood sampling and a proof of entitlement for sampled chunks.

The module has no tokio, storage, or node dependencies. It is a pure function of its inputs and reuses nectar's own `SwarmAddress`/`ChunkAddress`, `Proof`, and `Prover` (BMT) types.

## What it adds

- `canonical_neighbourhood(anchor, depth, addrs) -> Vec<ChunkAddress>`: filters addresses to those within `depth` of the anchor (proximity order at least `depth`) and returns them sorted by ascending **full 256-bit XOR distance** via `SwarmAddress::distance_cmp`. The full distance is used deliberately rather than `extended_proximity`, which caps at 36 bits: addresses sharing a 36-bit prefix would otherwise collapse into an equal, unstably ordered class and break cross-peer determinism.
- `sample(seed, candidates, n) -> Vec<usize>`: deterministic seed-driven selection of distinct indices (sampling without replacement via a Keccak-keyed partial Fisher-Yates). Identical on every node given the same seed and candidate set.
- `Entitlement { address, proof, seed }` with `build(seed, address, chunk_bytes, segment_index)` (generates a BMT inclusion proof through the existing `Prover`) and `verify(root_hash)` (delegates to `Proof::verify`).

## Motivation

The redistribution game requires every participating node to converge on the same sample of a neighbourhood for a given commitment round, and to present a verifiable witness that it stores the sampled chunk content. `canonical_neighbourhood` and `sample` provide the deterministic, uniform selection; `Entitlement` provides the proof of entitlement. Placing these in `nectar-primitives` keeps the determinism guarantees in one shared, dependency-light location that vertex's `RedistributionStore` can build on.

## Testing

- `canonical_neighbourhood`: determinism (independent of input ordering) and ascending full-distance sort order, plus depth filtering.
- `sample`: determinism (same seed yields the same indices), distinctness, bounds, full-permutation behaviour when `n` exceeds the candidate count, and the empty case.
- `Entitlement`: build and verify round-trip across several segment indices, including rejection of a wrong root.

`cargo test -p nectar-primitives`, `cargo clippy -p nectar-primitives --all-targets`, and `cargo fmt` all pass.

## AI-assistance disclosure

This contribution was prepared with AI assistance (Claude Code). All code, documentation, and test design were reviewed by the author prior to submission.

### DIFF
--- a/crates/primitives/src/lib.rs
+++ b/crates/primitives/src/lib.rs
@@ -45,6 +45,7 @@ pub mod network_id;
 pub mod nonce;
 pub mod overlay;
 pub mod proximity_order;
+pub mod redistribution;
 pub mod signing;
 pub mod spec;
 pub mod store;
@@ -70,6 +71,7 @@ pub use network_id::NetworkId;
 pub use nonce::Nonce;
 pub use overlay::compute_overlay;
 pub use proximity_order::{ProximityOrder, ProximityOrderError};
+pub use redistribution::{Entitlement, canonical_neighbourhood, sample};
 pub use spec::{MAINNET, StaticSpec, SwarmSpec, TESTNET};
 pub use timestamp::{Timestamp, TimestampError};
 
--- a/crates/primitives/src/redistribution.rs
+++ b/crates/primitives/src/redistribution.rs
@@ -1,0 +1,349 @@
+//! Redistribution-game primitives.
+//!
+//! Pure compute helpers underpinning the storage-incentives redistribution
+//! game: deterministic, uniform sampling over a neighbourhood and a proof of
+//! entitlement for an individual chunk. Everything here is a pure function of
+//! its inputs, with no I/O, no storage, no node, and no async machinery, so it
+//! produces identical results on every participating node given identical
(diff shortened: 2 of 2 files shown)


mfw78:
Closing — these redistribution primitives belong in vertex, not nectar.

nectar is the shared Swarm primitive library (chunks, BMT, stamps, addresses, proximity). The redistribution game (uniform neighbourhood sampling + proof of entitlement) is a storer-role concern: not all nodes participate in the storage-incentive game, so it should not be in every nectar consumer's dependency closure. It will be re-homed in vertex's `vertex-swarm-redistribution` crate, consuming nectar's existing BMT `Proof`/`Prover` and proximity primitives.

The implementation here is sound and will be ported as-is. Not deleting the branch yet (porting reference).

PR #99, opened 2026-06-17
feat(primitives): add anchor-keyed transformed address and SOC unwrap

mfw78:
## Summary

Adds the anchor-keyed *transformed address* as a nectar primitive, so that nectar owns the bee parity oracle for the redistribution sampler instead of vertex hand-rolling it. This is PART A of an agreed refactor of vertex PR #397 (redistribution sampler and proof of entitlement); PART B shrinks vertex to consume these primitives.

## Changes

- `AnyChunk::transformed_address(&self, anchor: &[u8]) -> ChunkAddress`, an inherent method alongside `address()`/`span()`/`data()`. It reproduces bee's `storer.transformedAddress` (`pkg/storer/sample.go`) byte-for-byte:
  - For a content-addressed chunk (CAC), the result is the prefixed BMT of the chunk body, keyed by the anchor via `Hasher::with_prefix` (the per-node prefix from #98).
  - For a single-owner chunk (SOC), the wrapped content chunk is re-hashed the same way to obtain `inner`, then the SOC transformed address is a plain, unprefixed `keccak256(soc_address || inner)`, mirroring `transformedAddressSOC`.
  - The CAC core is factored into a private helper that mirrors bee's `transformedAddressCAC`.
  - `span()`/`data()` already delegate to the inner BMT body for both CAC and SOC, so the SOC inner-CAC needs no `id`/`signature` header slicing.
- `SingleOwnerChunk::unwrap_cac(&self) -> ContentChunk` and `inner_body(&self) -> &BmtBody`, mirroring `soc.UnwrapCAC` (`pkg/soc/soc.go`), so callers stop slicing the wrapped CAC at a fixed cursor.

## Parity

New unit tests reproduce bee's published vectors:

- CAC: bee's `TestSampleVectorCAC` (4096-byte `i % 256` payload, anchor `swarm-test-anchor-deterministic!`). Plain BMT address `902406053a7a2f3a17f16097e1d0b4b6a4abeae6b84968f5503ae621f9522e16`, transformed address `9dee91d1ed794460474ffc942996bd713176731db4581a3c6470fe9862905a60`.
- SOC: a single-owner item from bee's `TestMakeInclusionProofsRegression` oracle (anchor `0x64`), transformed address `521f50f895dc1deea14448c09ba8d9c510c5db09cd84c7ed8413b66f15fbc110`.

Both reproduce byte-for-byte. `cargo test -p nectar-primitives`, `cargo clippy`, and `cargo fmt` are clean.

## AI disclosure

This change was prepared with the assistance of an AI coding agent (Claude). All code, vectors, and the bee cross-check were reviewed by the author before submission.

### DIFF
--- a/crates/primitives/src/chunk/any_chunk.rs
+++ b/crates/primitives/src/chunk/any_chunk.rs
@@ -3,9 +3,10 @@
 //! This module provides [`AnyChunk`], an enum that can hold any chunk type
 //! for runtime polymorphism without requiring trait objects.
 
+use alloy_primitives::Keccak256;
 use bytes::Bytes;
 
-use crate::bmt::DEFAULT_BODY_SIZE;
+use crate::bmt::{DEFAULT_BODY_SIZE, Hasher};
 use crate::error::Result;
 
 use super::chunk_type::ChunkType;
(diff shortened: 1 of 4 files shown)


crates/primitives/src/chunk/any_chunk.rs
```diff
@@ -111,6 +112,71 @@ impl<const BODY_SIZE: usize> AnyChunk<BODY_SIZE> {
...
+    ///
+    /// Mirrors bee's `transformedAddressCAC` (`pkg/storer/sample.go`): a BMT
+    /// keyed by `anchor` where the anchor prefixes every node hash. The span is
+    /// written little-endian by the hasher.
+    fn transformed_address_cac(anchor: &[u8], span: u64, payload: &[u8]) -> B256 {
+        let mut hasher: Hasher<BODY_SIZE> = Hasher::with_prefix(anchor);
+        hasher.set_span(span);
+        hasher.update(payload);
+        hasher.sum()
+    }
```
mfw78: why can't this just take the BmtBody, which would have the span and payload and it would also have the BODY_SIZE as well?

mfw78 (commented):
Should also consider the performance on this and see if borrowing can help as well.

crates/primitives/src/chunk/any_chunk.rs
mfw78: Done in 5c49661. The CAC core is now an inherent method on the typed body — `BmtBody::transformed_root(&self, anchor) -> B256` — so it reads the body's own `span()`/`data()` and carries `BODY_SIZE`, no raw `(span, payload)` args. `AnyChunk::transformed_address` dispatches through borrowed accessors (`ContentChunk::body()` / `SingleOwnerChunk::inner_body()`), cloning nothing; the private `transformed_address_cac` helper is gone. Bee CAC+SOC vectors still reproduce byte-for-byte.

mfw78:
On the performance/borrowing point: the same 5c49661 refactor addresses it — both standard paths now go through a borrowed `&BmtBody` (`body()`/`inner_body()`) and clone nothing. Related finding: `SingleOwnerChunk::unwrap_cac()` clones (`from_body(self.body.clone())`) and its only consumer in the workspace is the SOC vector test; with `inner_body()` giving a zero-copy `&BmtBody`, it is redundant for the transformed-address path. I left it in place in case it is intended API — happy to drop it if you prefer.

PR #105, opened 2026-06-25
feat(primitives): add concurrent out-of-order offset stream to the joiner

### LINKED ISSUE #106: Joiner has no concurrent out-of-order streaming read, forcing buffered or subtree-serialized reassembly
## Problem
The `Joiner` offers only ordered or fully-buffered reads: `into_stream` walks the top-level subtrees one at a time and emits leaves in order (so its effective concurrency is the number of top-level subtrees, not the configured width), and `read_range`/`read_all` fetch concurrently but buffer the whole result into a `Vec`. There is no mode that is concurrent, streaming, and out-of-order at once.

## Why it matters
A consumer cannot stream a large file with bounded memory at full concurrency. It must either buffer the whole file (heap blows up on large files) or accept subtree-serialized concurrency (deep trees ramp slowly and never reach the configured width), and a single slow leaf gates everything ordered behind it. This forced the vertex browser client to re-implement the tree walk by hand and is the root primitive behind vertex #521 (whole-file OOM), #522 (ordered head-of-line stall), and #523 (level-synchronous prefetch).

## Proposed approach
Add concurrent, out-of-order joiner consumption modes that yield each leaf tagged with its absolute byte offset the instant it lands: `into_offset_stream` (subtree-granular) and `into_offset_stream_chunked` (a chunk-granular pool that holds the configured width in flight regardless of subtree count). Consumers reassemble by offset and stay bounded by the in-flight width, not the file size.

## Status
A reference implementation is up as #105 (this issue is the backing problem statement that PR should close). A WASM-friendly retry path for the chunked stream additionally depends on #39 (ChunkGet without the Send bound).

## Related
- nectar #105 (the fix)
- nectar #39 (WASM ChunkGet without Send bound, needed for non-blocking retry-with-timeout)
- vertex epic nxm-rs/vertex#526 and issues nxm-rs/vertex#521, #522, #523 (the node/demo consumers)

AI Assistance: Claude Code used for the live-mainnet diagnosis of the vertex prototype that surfaced this primitive gap, and for drafting this issue.

mfw78:
## What

Add two concurrent out-of-order joiner consumption modes that yield each leaf body tagged with its absolute byte offset in the file, the instant it lands.

- `into_offset_stream`: keeps up to `concurrency` *subtree* fetches in flight and yields `(byte_offset, leaf_body)` pairs as each subtree resolves. The subtree walk is a sequential descent, so its effective leaf concurrency is the number of top-level subtrees, not the configured width.
- `into_offset_stream_chunked`: walks the tree at *chunk* granularity. Every intermediate and leaf fetch competes for one bounded in-flight pool, so the configured width of individual chunks is in flight regardless of how few top-level subtrees the tree has. A failed leaf fetch is re-enqueued onto the work queue up to a bounded retry budget, so a flaky chunk retries without holding its slot or gating the ready leaves behind it. The walk is a producer/worker model expressed without a spawned task, so it is Send-agnostic and runs single-threaded on wasm.

The joiner previously offered two modes, neither of which is concurrent, streaming, and out of order at once:

- `into_stream` walks the top-level subtrees one at a time and emits bytes strictly in file order (streaming, but sequential across subtrees).
- `read_range`/`read_all` fetch subtrees concurrently but buffer the whole result into a `Vec` (concurrent, but not streaming).

Both new modes reassemble by offset into a random-access sink, so a slow chunk never gates the leaves ready behind it, and peak memory is bounded by the in-flight width rather than the file size. The width honours `with_concurrency`. `into_stream`, `read_range`, and `read_all` are unchanged. The new paths are generic over the same `G: ChunkGet` / `M: JoinMode` / `BODY_SIZE` as the rest of the joiner and reuse the existing subtree frontier and per-leaf offset arithmetic; no node or libp2p dependency is introduced.

## Why

A streaming, bounded-memory download driven by the core joiner. A downstream consumer (the vertex swarm-demo browser node) currently bolts a separate concurrent prefetch ahead of the in-order `into_stream`, and separately re-implements the whole tree walk, offset arithmetic, and concurrency for its out-of-order path, because the core offered no concurrent out-of-order streaming mode.

Measuring the subtree-granular `into_offset_stream` from the browser against the live network showed it regressed throughput against the consumer's per-chunk prefetch path: for the typical two-level tree the effective leaf concurrency is the subtree count, so the configured width never bound the fan-out. `into_offset_stream_chunked` realises the configured width faithfully (the consumer's per-chunk retained-memory gauge tracks the width exactly), recovering the median throughput while keeping memory bounded.

No existing issue tracks this; filing the primitive directly.

## Testing

`cargo test -p nectar-primitives --features tokio` (and with `encryption`). All 223 unit tests pass; `cargo clippy -p nectar-primitives --all-targets --all-features -- -D warnings` is clean.

New tests in `joiner.rs` assert the core invariant for both modes: the offset stream yields every leaf exactly once, the covered byte count equals the file size, and reassembling the pairs by offset is byte-for-byte equal to both `read_all()` and the original input. Covered: sub-chunk, exact-chunk, multi-chunk-with-tail, three-level tree, and a concurrency-of-1 degenerate fan. The chunked mode additionally has a per-chunk concurrency proof: a gating getter on a flat two-level tree (one wide subtree) observes the in-flight fetch count reaching the configured width, which the subtree-granular walk cannot reach.

## AI Assistance

AI Assistance: Claude Code used for the implementation, tests, the throughput measurement against the live network, and this description.

Closes #106

### DIFF
--- a/crates/primitives/src/file/joiner.rs
+++ b/crates/primitives/src/file/joiner.rs
@@ -289,6 +289,77 @@ where
         })
     }
 
+    /// Convert into a stream of `(byte_offset, leaf_body)` pairs fetched with
+    /// bounded concurrency and yielded out of order as each leaf lands.
+    ///
+    /// Unlike [`into_stream`](Self::into_stream), which walks subtrees one at a
+    /// time and emits bytes in file order, this keeps up to `concurrency`
+    /// subtree fetches in flight and yields each leaf the moment its subtree
+    /// resolves, tagged with its absolute byte offset in the file. Reassembling
+    /// the pairs by offset reproduces the file; nothing is buffered into a
+    /// single contiguous result, so peak memory is bounded by the in-flight
+    /// width, not the file size. Set the width with
+    /// [`with_concurrency`](Self::with_concurrency).
+    pub fn into_offset_stream(self) -> impl Stream<Item = Result<(u64, Bytes)>> {
+        let getter = self.getter;
+        let concurrency = self.concurrency;
+        let chunk_range = self.tree.chunks_for_range(0, self.span);
+
+        // Each subtree fetch resolves to its base offset plus its leaves in tree
+        // order; `buffer_unordered` yields whichever subtree finishes first.
+        let subtrees = stream::iter(self.subtrees)
+            .map(move |st| {
+                let getter = Arc::clone(&getter);
+                async move {
+                    let base = st.byte_offset;
+                    let bodies =
+                        read_subtree_bodies_async::<G, M, BODY_SIZE>(&*getter, &st, &chunk_range)
+                            .await?;
+                    Ok::<(u64, Vec<Bytes>), FileError>((base, bodies))
+                }
+            })
+            .buffer_unordered(concurrency.max(1));
+
+        // Flatten each resolved subtree into per-leaf `(offset, body)` pairs,
+        // draining one subtree's leaves before polling the next ready subtree.
+        struct State<S> {
+            subtrees: S,
+            base: u64,
+            leaf_index: usize,
+            pending: std::vec::IntoIter<Bytes>,
+        }
+
+        let state = State {
+            subtrees: Box::pin(subtrees),
+            base: 0,
+            leaf_index: 0,
+            pending: Vec::new().into_iter(),
+        };
+
+        stream::unfold(state, move |mut state| async move {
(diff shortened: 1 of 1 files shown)


mfw78:
Superseded by #110

PR #222, opened 2026-07-11
chore(lint): exempt non-shipping targets from restriction lints

mfw78:
## What

Exempts non-shipping compilation targets from the workspace restriction lints: benches, examples, and integration tests get a file-level allow header stating the policy (unwraps, direct indexing, casts, and assertions there are setup and illustration, not shipped surface). Two shipped-code follow-ups ride along: a `saturating_sub` in the postage generator (provably identical, the value is always at least 1) and two provably in-bounds indexing allows in the mantaray arbitrary impl.

## Why

The restriction lints protect shipped library code, but benches, examples, and integration tests are standalone compilation units the `cfg(test)` exemption never covered; under the full battery (`cargo clippy --workspace --all-targets --all-features -- -D warnings`) they produced several hundred errors that made the strict profile unusable as a gate. With this car the full battery is clean at the train tip.

## Testing

`cargo clippy --workspace --all-targets --all-features -- -D warnings` exits clean at this commit; the narrow CI profile is unaffected.

## AI Assistance

Authored by Claude Fable under maintainer direction during train conducting.

### DIFF
--- a/crates/mantaray/benches/mantaray_bench.rs
+++ b/crates/mantaray/benches/mantaray_bench.rs
@@ -1,5 +1,16 @@
 #![allow(missing_docs)]
-
+// Bench, example, and integration-test code: unwraps, direct indexing,
+// casts, and assertions are setup and illustration, not shipped surface.
+#![allow(
+    clippy::unwrap_used,
+    clippy::expect_used,
+    clippy::indexing_slicing,
+    clippy::arithmetic_side_effects,
+    clippy::panic,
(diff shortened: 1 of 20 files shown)


mfw78:
The per-file headers here are interim: the dev-crate consolidation that deletes all 18 of them is now specced (see the new workspace refactor issue) - per-target lint scoping does not exist in cargo (RFC 3389 deferred) and the clippy.toml test keys are broken for standalone targets (rust-clippy#13981).

Issue #230, opened 2026-07-11
refactor(lint)!: burn down restriction-lint suppressions to near zero in shipped src


mfw78:
The restriction-lint train set deny-level lints and then suppressed its way past them: the as_conversions car alone added 60 `#[allow(clippy::as_conversions)]` attributes against 74 genuinely refactored conversions, and shipped-src suppressions grew from 179 to 233 across the trains. Justification comments do not change the arithmetic: the lints exist to force tighter code, and an allow with a proof-comment is still unchecked prose where the compiler could hold a proof.

**Mandate: refactor first. A suppression is the documented exception, not the response.**

Burn-down, per lint:

- `as_conversions`: every widening cast becomes `From`/`usize::from`/`u64::from`; every narrowing becomes `TryFrom` with a real error path, or a typed accessor that returns the narrow type at the source so no cast exists at the use site. Const contexts, where `From` is not const-callable, get one named const helper per conversion shape (for example `const fn widen_u8(x: u8) -> usize`) rather than a per-site allow. The helper is then the single audited cast.
- `indexing_slicing`: replace checked-by-comment indexing with slice patterns, `split_first`/`split_at_checked`, `get` plus an error, iterators, or fixed-size array destructuring. The `wire::Cursor` rewrite already proved this class refactors away cleanly; apply the same treatment to the survivors, which are the encode paths, prefix handling and the fork cursor.
- `arithmetic_side_effects`: bounded arithmetic gets the bound INTO the types, through `checked_`/`saturating_` operations with the invariant asserted once at construction, or through width-limited newtypes. A comment asserting the bound holds is not enough.
- `unwrap_used`/`expect_used` in src: restructure so the invariant is type-carried, through typestate, `NonZero` or exhaustive enums. An expect that survives must be on an invariant the type system genuinely cannot express.

Acceptance:

1. Shipped src (`crates/*/src` outside `cfg(test)`) carries near-zero suppressions. Each survivor is `#[expect(clippy::..., reason = "...")]`, which is self-cleaning when stale, and names an invariant that provably cannot be typed or restructured away. Each is reviewed case by case.
2. `clippy::allow_attributes` is denied workspace-wide so a bare `#[allow]` cannot return. It is not denied today: `[workspace.lints.clippy]` in the root `Cargo.toml` sets the four restriction lints above but no `allow_attributes`.
3. The suppression count is part of review: any pull request that raises it must say so in its body.
[...]

PR #232, opened 2026-07-12
chore(lint): deny clippy::as_conversions and make all casts explicit

### LINKED ISSUE #151: Deny clippy::as_conversions (lossy casts) workspace-wide
### Goal (maximal tier, follow-up to #149)

Promote `clippy::as_conversions` from `warn` to `deny` in `[workspace.lints.clippy]` and eliminate the remaining roughly 199 lossy/`as`-cast sites so all casts are explicit and checked.

### Impact

Silent lossy truncation on decode and encode paths is currently possible wherever an `as` cast narrows an integer. Denying the lint forces each cast to be either a provably-lossless widening, a checked `try_from` with a typed error, or an explicitly justified truncation, removing that class of bug.

### Strategy (behaviour-preserving)

- Safe widening (`u8 as usize`, `u32 as u64`, and similar): replace with `usize::from(x)` / `u64::from(x)` (byte-identical, clearer intent).
- Untrusted narrowing: replace with `try_into()` and a typed error on the parse/decode paths.
- Provably-safe intentional truncation/wrap (for example `ref_size as u8` where `ref_size` is in {32, 64}): justified `#[allow(clippy::as_conversions)]` with the invariant stated, or `TryFrom` with expect-free handling.

Tests are exempted via the existing crate-root `#![cfg_attr(test, allow(...))]`.

### Distribution (post-#149)

primitives ~107, postage-usage ~39, postage-issuer ~29, mantaray ~25, swarms ~12, postage ~5.

Stacked on the aggressive lint PR (#150).

Resolved by #152.

mfw78:
## What

Flip `clippy::as_conversions` from allow to `deny` workspace-wide and resolve every resulting site (roughly 150) across `nectar-primitives`, `nectar-postage`, `nectar-swarms`, `nectar-mantaray`, `nectar-postage-issuer`, and `nectar-postage-usage`.

Behaviour-preserving: no public signatures, error types, or wire/encoding formats changed.

The config flip adds `clippy::as_conversions = "deny"` to the workspace lint table. The remaining commits fix casts one crate at a time.

### Per-crate cast fixes (from-widenings / try_from-narrowings / justified-allows)

| crate | from | try_from | allow | tests |
|---|---|---|---|---|
| nectar-primitives | 67 | 0 | 16 | 202 lib + 16 doctest (287+17 all-features) |
| nectar-postage | 0 | 0 | 4 | 47 unit + 1 doctest |
| nectar-swarms | 3 | 1 | 3 | 20 |
| nectar-mantaray | 14 | 1 | 7 | 72 unit + 4 integration + 3 doctest |
| nectar-postage-issuer | 5 | 0 | 24 | 52 unit + 1 doctest |
| nectar-postage-usage | 22 | 0 | 21 | 27 unit + 29 snapshot + 3 vector + 7 doctest |

### Strategy (behaviour-preserving)

1. Pure widening (`u8/u16/u32/usize` to `u64/usize`, provably lossless on every target): replaced with `T::from(x)`. No behaviour change, no new error path.
2. Narrowing on untrusted/attacker-controlled input where a rejection path made sense: replaced with `T::try_from(x)` mapped to the pre-existing error type, verified to preserve the exact accept/reject boundary and error payload. For example, `nectar-swarms` serde `visit_i64` now uses `u64::try_from(v).map_err(...)`, which accepts/rejects identically to the old `v.is_negative()` check; `nectar-mantaray` `Fork::encode_into` metadata-length guard now uses `u16::try_from` into the existing `MetadataTooLarge` error.
3. Provably-safe truncation (value bounded by prior validation, `const fn` where `T::from`/`try_from` is not const-callable, or truncation is the intended semantic such as extracting low bits): narrow, comment-justified `#[allow(clippy::as_conversions)]` at the expression or function level, never crate-wide.
4. Where the original cast already had no lossless equivalent and no sensible rejection path (for example `nectar-primitives::cast::usize_from_u64`, used on decode-path spans with no pre-existing error branch), the exact original truncation semantics, including pre-existing wasm32 truncation behaviour, were preserved verbatim via a documented crate-internal helper rather than introducing a new fallible path. Introducing one would have been a behaviour change, not a hardening.
5. A handful of `&mut T as *mut T` trait-object/pointer coercions were replaced with `std::ptr::from_mut` or explicit turbofish typing (byte-identical, not lossy casts, done for lint-cleanliness only).

No production `as` casts on untrusted decode paths were left un-justified; the two `try_from` red-team sites above were specifically flagged for adversarial review and came back clean (see Testing).

## Why

Closes #151.

Stacked on #150 (`fuzz/5-clippy-hardening`). Denying `as_conversions` forces every integer cast to be explicit and checked, removing a class of silent lossy-truncation bugs on decode and encode paths.

## Testing

Full-workspace gate:

```bash
cargo clippy --workspace --locked   # exit 0, zero warnings/errors, as_conversions deny holds
cargo test --workspace              # exit 0, all suites + doctests green
git status --porcelain              # empty, working tree clean
```

Per-crate spot checks used during implementation:

```bash
cargo clippy -p nectar-<crate> [--no-deps] [--all-features] [--all-targets]
cargo test -p nectar-<crate> [--all-features]
```

Observed: `cargo clippy --workspace --locked` exits 0 with the `deny` in place, and `cargo test --workspace` passes all listed unit, integration, doctest, snapshot, and vector suites.

## Known limits

- `benches/` and `tests/` targets are outside the enforced lint surface (CI runs `cargo clippy --workspace --locked` against lib targets only); some pre-existing `unwrap_used`/`indexing_slicing` diagnostics remain there under `--all-targets`, unrelated to this change and out of scope.
- `nectar-swarms/src/lib.rs` is missing the `#![cfg_attr(test, allow(...))]` exemption block present in postage/mantaray. Pre-existing, not introduced or fixed here, does not gate CI.
- One unrelated pre-existing `fmt` drift in `chunk_type_set.rs` was intentionally left untouched (reverted from an incidental fix) to keep the primitives change focused.
- Each crate was individually verified as clippy-clean and green before the dependent-crate compile issue (primitives blocking downstream `cargo clippy` without `--no-deps`) was resolved by the final `nectar-primitives` change; the workspace `cargo clippy --workspace --locked` run above confirms the full stack is clean end to end.

AI Assistance: Claude (Anthropic) used for the implementation, tests, and this PR description.

---
Supersedes #152, which was auto-closed and locked by GitHub after a botched force-push briefly zeroed its diff during a stack rebase; branch content is identical and tree-verified.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,8 +41,9 @@ unreachable = "deny"
 todo = "deny"
 unimplemented = "deny"
 panic_in_result_fn = "deny"
-# Lossy/`as` casts: warn now; denied in the maximal follow-up PR.
-as_conversions = "warn"
+# Lossy/`as` casts: all casts must be explicit (`from`/`try_into`) or a
+# justified `#[allow]` where truncation is intended and provably in range.
+as_conversions = "deny"
 missing_panics_doc = "warn"
 
 [workspace.metadata.docs.rs]
--- a/crates/mantaray/src/codec.rs
+++ b/crates/mantaray/src/codec.rs
@@ -123,7 +123,9 @@ fn encode_node<E: NodeEntry>(node: &Node<E>) -> Result<Vec<u8>> {
     data[NodeHeader::VERSION_HASH_OFFSET..NodeHeader::VERSION_HASH_OFFSET + VersionHash::SIZE]
         .copy_from_slice(VersionHash::V02.as_bytes());
 
-    data[NodeHeader::REF_SIZE_OFFSET] = ref_size as u8;
+    #[allow(clippy::as_conversions)] // ref_size = E::SIZE (32 or 64), always fits u8
+    let ref_size_byte = ref_size as u8;
+    data[NodeHeader::REF_SIZE_OFFSET] = ref_size_byte;
 
     // append entry (or E::SIZE zero bytes if empty)
     match &node.entry {
@@ -134,7 +136,7 @@ fn encode_node<E: NodeEntry>(node: &Node<E>) -> Result<Vec<u8>> {
     // build the 256-bit index of which fork bytes are present
     let mut index = U256::ZERO;
     for &fork_byte in node.forks.keys() {
-        index.set_bit(fork_byte as usize, true);
+        index.set_bit(usize::from(fork_byte), true);
     }
     data.extend_from_slice(&index.to_le_bytes::<32>());
 
@@ -241,7 +243,7 @@ fn decode_empty_terminal_node<E: NodeEntry>(data: &[u8]) -> Result<Node<E>> {
 // (<= 64) that cannot overflow usize.
 #[allow(clippy::indexing_slicing, clippy::arithmetic_side_effects)]
 fn decode_v01<E: NodeEntry>(data: &[u8]) -> Result<Node<E>> {
-    let ref_bytes_size = data[NodeHeader::REF_SIZE_OFFSET] as usize;
+    let ref_bytes_size = usize::from(data[NodeHeader::REF_SIZE_OFFSET]);
     // BEE-WORKAROUND(bee#5483): see HAZMAT block above `decode_empty_terminal_node`.
     if ref_bytes_size == 0 {
         return decode_empty_terminal_node::<E>(data);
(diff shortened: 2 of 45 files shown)


mfw78:
Review note: this car added 60 as_conversions suppressions vs 74 refactored conversions; #230 (rewritten) now mandates the refactor-first burn-down of these after the trains land.

Issue #264, opened 2026-07-12
epic: implement mantaray 1.0 as a key-value store (nectar-manifest crate)


mfw78:
Implement the frozen mantaray 1.0 design (Draft v3 spec + paper) as a new **nectar-manifest** crate: a clean-slate, content-addressed **key-value store** whose folder/website behaviour is a **view** over the KV core. This is a redesign, not a revision of 0.2 - 0.2 stays readable by 0.2 readers, coexisting on the network.

## Architecture

A compacted radix-256 trie over arbitrary byte keys, stored as CAC chunks (address = BMT of the node payload; an in-payload `magic 0x6D || version` prefix, no chunk-type registration). Content-defined packing (CDC) makes the tree a pure function of its key set - canonical bytes, cross-version dedup, and history-independent updates. Values and typed metadata ride inline in fork records (no leaf-chunk indirection). Built on the 0.6.0 typed chunk model (`Chunk<Verified>`, `ChunkRegistry`, trusted store traits).

## Cross-cutting constraints (hard, every car)

- **Type-carried constants.** The frozen tag_version 0x01 parameters live as associated consts on a sealed `Format` trait (INLINE_MAX=1536, SEG_TARGET=2048, CUT_SCALE=2^53, SEG_MIN=512, BUDGET=4094, PLEN_MAX=255, FORKS_MAX=256, ...). Codec/trie generic over `F: Format`, public API defaults to `V1` so users never see the generic. Bounded newtypes (`Prefix<F>`, MetadataLen, ...) carry bounds as type invariants. **No magic-number free consts scattered in code.**
- **Codec discipline.** The only fallible byte access is `wire::Cursor`. **No hand-summed slicing, no unwrap/expect, no `#[allow]`** for indexing/unwrap - refactor-first; an unavoidable suppression is `#[expect(reason)]` for a genuinely untypeable invariant only.
- **Memory-bounded / streaming.** Never hold the whole tree; peak memory O(depth) not O(tree). Runs on resource-constrained devices (phones, desktops). **wasm in scope, no_std out of scope.**
- **UX/DX paramount.** It should be second nature to take a bunch of files, upload via BMT, assemble into a manifest, and publish. Terse docs (intent + the one non-obvious invariant).

## Sequencing

[...]

PR #307, opened 2026-07-13
feat(manifest): weave order-statistic counts into the baseline wire grammar

### LINKED ISSUE #296: design+feat(manifest): order-statistic subtree counts for O(depth) range-count and pagination
To make `count(lo, hi)` and offset/limit pagination ("items 1000-1050 of this listing") cost O(depth) instead of O(window), persist a per-fork **subtree key-count** (an order-statistic / counted-trie annotation). Navigation by rank then never fetches the window's leaves. This is the classic counted-B-tree trick and is a large UX win for any paginated view.

**Format-affecting** - this changes node canonical bytes, so it MUST NOT mutate the frozen `tag_version 0x01`.

## Phase 1 - design spike (do first, gate before Phase 2)
Recommend the carrier, with a short written rationale:
- (a) a new `tag_version 0x02` / mantaray 1.1 (counts are part of the node grammar), or
- (b) a `Format`-gated optional field / well-known reserved node-metadata entry (counts-on vs counts-off as a build mode producing distinct addresses), or
- (c) defer.
Requirement for any carrier: the count is a pure deterministic function of the subtree, so history-independence (I6) + dedup still hold (a counts-on manifest is canonical among counts-on manifests). Surface the recommendation to the maintainer before committing to a new tag_version.

## Phase 2 - implement behind the chosen carrier
- Persist subtree counts, maintained bottom-up by the builder and by `apply` (derivable at seal time; keep it I6-safe).
- O(depth) `count(lo, hi)`, rank/select, and `paginate(prefix/range, offset, limit)` APIs.
- Conformance: rank/count agree with a full-walk oracle; canonical bytes stable under shuffled build order.

Depends on the packing / builder / apply / read-ops cars. Part of the range-query performance set under epic #264.
Cross-cutting (see epic): type-carried Format consts (no free consts); `wire::Cursor` only fallible byte access (no slicing/unwrap/#[allow]); terse docs; memory-bounded; wasm in scope.

mfw78:
## What

Makes the frozen `tag_version 0x01` (`V1`) grammar itself carry order-statistic
subtree counts, scoped entirely to `crates/manifest`, for #296 (epic #264 Phase
2). There is no longer a separate counted profile: the baseline wire is
counted, `V1Read` (`0x02`) inherits it as before, and the earlier `V1Counted`
(`0x03`) profile is gone.

- `count.rs`: unchanged in role — a `SubtreeCount` newtype with a canonical
  minimal-length uleb128 codec over `wire::Cursor`/`Writer`. The decoder
  rejects overlong (non-minimal) encodings and over-wide runs via checked
  shifts, is bounded by a 10-byte (`MAX_WIRE_BYTES`) cap, and contains no
  slicing, `unwrap`, or `#[allow]`. A public `CountError` enum (`Underrun`,
  `Overlong`, `TooWide`) reports rejections.
- `format.rs`: the `Format` trait loses the `COUNTED` discriminant and
  `V1Counted` is deleted. `V1`'s fork-record termination bound is re-derived
  to add the worst-case varint count weight unconditionally, and `V1Read`
  inherits it by reusing `V1`'s layout constants as before; both are
  re-asserted via `assert_layout!`.
- `codec.rs`: fork records unconditionally gain a trailing `child_count`
  uleb128 when the child is referenced (`Ref32`/`Ref64`); it is omitted for
  embedded/leaf children. Segment-directory descriptors unconditionally gain
  an analogous `seg_count`. There is no longer a const-folded uncounted path.
- `apply.rs`, `builder.rs`, `fork.rs`, `packing.rs`: thread `SubtreeCount`
  through verbatim-reuse, chain, compact, reroot and split so `apply(root,
  delta)` reproduces manifests bit-exactly, since a subtree's count is a pure
  function of its key set (`(entry?1:0) + sum(child counts)`). Directory
  packing weight now charges the worst-case `seg_count` width per descriptor
  as well, so a directory routing many maximal counts cannot overrun its
  segment chunk.
- `tests/counted.rs`: integration coverage for the baseline (round-trip,
  canonicality under shuffled build order, apply-vs-rebuild). Renamed off the
  `_with_counts` naming this rework makes redundant, since counts are no
  longer an optional add-on: `build_then_read_round_trips_every_key`,
  `apply_matches_a_from_scratch_build`, `an_inline_value_round_trips`.
- Existing fixture/vector tests were regenerated against the new baseline
  bytes.

The `0x01` version byte keeps its value but changes wire meaning: uncounted
payloads written by an earlier unreleased build no longer decode.

## Why

Order-statistic subtree counts let a manifest answer "how many keys in this
subtree" and support rank-based navigation in O(depth) rather than O(window).
Carrying the count on the baseline wire instead of behind a separate profile
removes a whole axis of format proliferation (no `V1`-vs-`V1Counted` split to
keep in sync) without disturbing anything else `V1` guarantees: the count is a
pure function of the key set, so canonical bytes and bit-exact `apply` hold
exactly as before. `V1Read` (`0x02`) continues to be a distinct wire version
carrying this same counted baseline with a heavier embedding budget.

Closes #296

## Testing

Full workspace battery, clean at the branch tip via `nix develop`:

- `cargo fmt --all --check`: clean.
- `cargo nextest run -p nectar-manifest -p nectar-manifest-proof --locked`:
  234 tests run, 234 passed, 0 skipped.
- `cargo nextest run --workspace --locked`: 837 tests run, 837 passed, 1
  skipped.
- `cargo test --doc --workspace --locked`: clean.
- `cargo fuzz build` plus `manifest_node_decode` / `manifest_build_roundtrip`
  / `manifest_apply_rebuild` targets (10000 runs each) and all committed seed
  corpora replay clean.
- A stale-reference sweep at the tip (`V1Counted`, `Format::COUNTED`,
  `uncounted`, `never emitted`, `unemitted`, `under the counted grammar`,
  `tag_version 0x03`) returns zero hits across `crates/`, `contracts/`, and
  `fuzz/`.

## AI Assistance

This PR's implementation, red-teaming and this description were produced with assistance from Claude (Anthropic), per the process in [nxm-rs/.github CONTRIBUTING.md](https://github.com/nxm-rs/.github/blob/main/CONTRIBUTING.md).

### DIFF
--- a/crates/manifest/src/apply.rs
+++ b/crates/manifest/src/apply.rs
@@ -23,6 +23,7 @@ use nectar_primitives::store::{ChunkPut, MaybeSync};
 
 use crate::bounded::Prefix;
 use crate::builder::{BuildError, BuildStats, Item, build_table, emit_node, resolve};
+use crate::count::SubtreeCount;
 use crate::error::{ForkPrefixEmpty, PrefixTooLong};
 use crate::fork::{Child, ForkPayload, ForkRecord, ForkTable};
 use crate::format::{Format, V1};
@@ -335,13 +336,14 @@ where
     }
 
     if deeper.is_empty() {
-        // The child is untouched: reuse it verbatim.
+        // The child is untouched: reuse it verbatim, carrying its stored count.
         return finish(
             store,
             edge,
             new_entry,
             new_meta,
             existing.child().cloned(),
+            existing.child_count(),
             stats,
         )
         .await;
@@ -353,7 +355,7 @@ where
             if items.is_empty() {
                 // A deletion of an absent deeper key: the fork is unchanged bar
                 // its terminal value.
-                return finish(store, edge, new_entry, new_meta, None, stats).await;
+                return finish(store, edge, new_entry, new_meta, None, None, stats).await;
             }
             build_table(store, &items, plen, stats).await?
         }
@@ -394,7 +396,7 @@ where
     F: Format,
 {
     if table.is_empty() {
-        return finish(store, edge, entry, meta, None, stats).await;
+        return finish(store, edge, entry, meta, None, None, stats).await;
     }
     // Edge-compaction: a child-only fork over a single-fork child merges into
     // one edge, exactly as a from-scratch build would compact the shared run.
@@ -404,8 +406,10 @@ where
     {
         return compact(store, edge, first, record, stats).await;
     }
-    let child = resolve(store, table, stats).await?.into_child();
-    finish(store, edge, entry, meta, Some(child), stats).await
+    let resolved = resolve(store, table, stats).await?;
+    let count = resolved.child_count();
+    let child = resolved.into_child();
+    finish(store, edge, entry, meta, Some(child), count, stats).await
 }
 
 /// An insertion diverges within the edge: branch at the divergence, re-rooting
@@ -485,6 +489,7 @@ where
         remainder,
         existing.payload().clone(),
         existing.metadata().cloned(),
+        existing.child_count(),
     )
 }
 
@@ -500,6 +505,7 @@ async fn finish<S, F>(
     entry: Option<Entry<F>>,
     meta: Option<Metadata<F>>,
     child: Option<Child<F>>,
+    child_count: Option<SubtreeCount>,
     stats: &mut BuildStats,
 ) -> Result<Option<ForkRecord<F>>, ApplyError>
 where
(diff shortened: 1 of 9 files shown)


mfw78:
Resolved the wire-version-byte collision flagged in review: `V1Counted` was sharing `VERSION = 0x02` with `V1Read`. Bumped `V1Counted` to `0x03` (empty-root prefix `6D 02` -> `6D 03`), leaving `V1` (`0x01`) and `V1Read` (`0x02`) untouched.

Added a const-block wire-version registry (`0x01` V1, `0x02` V1Read, `0x03` V1Counted) asserting all three version bytes are pairwise distinct, so a future silent collision becomes a build error. fmt/clippy/test all green; the child reader branch (#308-equivalent, s264/82-counted-reader) was rebased onto the fix and re-gated green.

New head: `89253b5` (signature: valid).

PR #309, opened 2026-07-13
feat(manifest-proof): add inclusion/exclusion proof crate

### LINKED ISSUE #301: feat(zkvm): manifest inclusion and exclusion proving in a zkvm guest
Deferred to Post v1 under epic #693. Rescoped 2026-07-18: the in-crate proof generators and Rust verifiers (PR #309) were closed in favour of zkVM guest proving, per the no_std and zkVM architecture decision.

## What

Build the guest foundations for proving manifest statements inside a zkVM.

- A `no_std` guest program that opens a manifest root and verifies inclusion (key has entry E under root R) and exclusion (key absent under root R) by walking node chunks with the ldb reader itself. The library is the verifier, so there is no bespoke proof format.
- Keccak routes through the patchable seam (alloy `keccak256`), accelerated in-guest through the sp1-patches and RISC Zero forks of the standard crates.
- Host-side harness: witness assembly, which is the node chunks on the path, guest invocation, and proof output for SP1 and RISC Zero.
- Cycle counts recorded per statement, as the baseline for the later aggregate work.

## Scope boundary, settled by the restructure plan

Two proof kinds are in scope across the programme: a segment of a chunk, which already exists, and membership or exclusion of a key in an ldb store, which is this issue.

A proof over a section of a file is out of scope. It has no stated consumer, and its interesting case is a byte range, which is a frontier rather than a path.

mantaray is excluded permanently. A mantaray root differs for the same content depending on insertion order, so it cannot anchor a claim about content, and it will never implement the proof seam.

## Correction: what the substrate now provides

The walker is `nectar-ldb`, not `nectar-manifest`. #641 renamed the key-value crate to `nectar-ldb`, and the `nectar-manifest` name belongs to the format-agnostic trait seam carved by #644.

Three moves in the restructure change the crates this issue links, so read the target names from `docs/RESTRUCTURE-PLAN.md` rather than from the text above.

`nectar-proof` is created in M2 by epic #800 and holds the authentication layer only: how one step's bytes bind to a trusted address, plus the replay loop. It carries no key vocabulary, no verdict and no span arithmetic. `Present` and `Absent` stay in `nectar-ldb-core`. So the trait plumbing this issue would otherwise have invented is provided; what remains here is the guest program, the host harness and the cycle counts.

`nectar-ldb` splits in M2 into `nectar-ldb-core`, which is `no_std` and carries node decode, descent verification and the proof implementation, and `nectar-ldb` above it, which is the std engine. A guest links `nectar-ldb-core`.

The proving core is renamed. Today's `nectar-primitives-core`, carved by #614 and landed by #642, becomes `nectar-primitives` in M2, and today's `nectar-primitives` dissolves.

The `no_std` precondition holds. `nectar-ldb` is `#![no_std]` (`crates/ldb/src/lib.rs`) and the bare-metal `riscv64imac-unknown-none-elf` lane in `.github/workflows/nostd.yml` gates it. Note that M0 narrows that matrix to the six proving-lane crates and no others, which are `nectar-primitives`, `nectar-spec`, `nectar-postage-primitives`, `nectar-proof` and `nectar-ldb-core`, plus `nectar-marker`. Anything this guest links must be on that list.

mfw78:
## What

Adds a new workspace crate, `nectar-manifest-proof`, providing a proof generator and verifier for inclusion and exclusion proofs over a mantaray 1.0 manifest, authenticated against a trusted root (BMT) address.

A `ForkPathProof` is an ordered chain of authenticated nodes from root to terminal.
Each step authenticates its node against the reference the previous node yielded, so the whole path forms a BMT hash chain anchored at the root.

`prove_inclusion` terminates at the target key's entry.
`prove_exclusion` terminates at the point where the descent provably cannot continue to the key, whether that is the absence of a fork for the next byte, a diverging compacted edge, or the key exhausting with no terminal entry present.

`verify(root, key, proof) -> Verdict::{Present(Entry) | Absent}` independently replays the descent over the authenticated bytes in the proof and returns whatever that replay yields, so a tampered proof, a wrong value, or a wrong root fails authentication rather than being mis-verified as present or absent.

New crate layout: `Cargo.toml`, `README.md`, `src/{lib,error,proof,descent,prove,verify}.rs`, `tests/roundtrip.rs`, plus `Cargo.toml`/`Cargo.lock` workspace wiring.

## Why

Closes #301

Manifest consumers need a way to prove and independently verify presence or absence of a key in a mantaray 1.0 manifest without trusting the party that produced the proof, given only the manifest's BMT root.

## Testing

Full battery run via `nix develop --command` (`cargo-nextest` supplied ephemerally via `nix shell nixpkgs#cargo-nextest`):

- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean, new crate compiles with no warnings.
- `cargo nextest run --workspace --all-features`: 937 passed, 0 failed, 1 skipped. The crate's proptest oracle test (`proofs_agree_with_the_reader`) runs approximately 60s and is flagged slow but passes.
- `cargo test --doc --workspace`.

An independent adversarial review found the descent parser to be byte-for-byte compatible with the manifest wire codec (verified field-by-field, including embedded children, ref64, and the V1Counted trailing-count layout), the multi-step authentication chain and gap-proof coverage to be sound, and confirmed empirically that truncated or adversarial input (oversized/garbage/empty proofs, intermediate-step tamper on a real two-node proof) cannot panic, owing to consistent use of a checked cursor and saturating/checked arithmetic. No correctness or soundness bugs were found.

This is a stacked PR targeting `s264/82-counted-reader`; no independent CI runs against this base until the stack is retargeted onto the trunk.

## AI Assistance

Claude (Anthropic) was used for code generation of the crate implementation and tests, for the independent adversarial review, and for this PR description.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2071,6 +2071,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "nectar-manifest-proof"
+version = "0.4.0"
+dependencies = [
+ "alloy-primitives",
+ "bytes",
+ "futures",
+ "nectar-manifest",
+ "nectar-primitives",
+ "proptest",
+ "thiserror",
+]
+
+[[package]]
 name = "nectar-mantaray"
 version = "0.4.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ check.all_targets = true
 # nectar crates
 nectar-contracts = { version = "0.4.0", path = "crates/contracts" }
 nectar-manifest = { version = "0.4.0", path = "crates/manifest" }
+nectar-manifest-proof = { version = "0.4.0", path = "crates/manifest-proof" }
 nectar-mantaray = { version = "0.4.0", path = "crates/mantaray" }
 nectar-primitives = { version = "0.4.0", path = "crates/primitives" }
 nectar-swarms = { version = "0.4.0", path = "crates/swarms" }
--- a/crates/manifest-proof/Cargo.toml
+++ b/crates/manifest-proof/Cargo.toml
@@ -1,0 +1,42 @@
+[package]
+name = "nectar-manifest-proof"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+exclude.workspace = true
+description = "Inclusion and exclusion proofs over a mantaray 1.0 manifest, verified against a trusted root"
+documentation = "https://docs.rs/nectar-manifest-proof"
+readme = "README.md"
+keywords = ["swarm", "ethereum", "manifest", "mantaray", "proof"]
+categories = ["data-structures", "cryptography"]
+
+[lints]
+workspace = true
+
+[dependencies]
+alloy-primitives = { workspace = true }
+nectar-manifest = { workspace = true }
+nectar-primitives = { workspace = true }
+thiserror.workspace = true
+
+[dev-dependencies]
+bytes.workspace = true
+futures.workspace = true
+proptest.workspace = true
+
+[features]
+default = [ "std" ]
+std = [
+	"alloy-primitives/std",
+	"nectar-manifest/std",
+	"nectar-primitives/std",
+	"thiserror/std",
+]
+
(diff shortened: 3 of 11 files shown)


mfw78:

PR #310, opened 2026-07-13
feat(manifest-proof): range-completeness and state-transition proofs

### LINKED ISSUE #302: feat(zkvm): aggregate manifest statements as guest programs
Deferred to Post v1 under epic #693. Rescoped 2026-07-18: the composed proof objects (PR #310) were closed in favour of zkVM guest proving.

## What

Aggregate manifest statements as guest programs, which is the workload class zkVMs are chosen for.

- Range-completeness: the entries between `lo` and `hi` under root R are exactly this set.
- State transition: root R2 is root R1 with changeset C applied, so the guest runs `apply` and checks the resulting root.
- Batch inclusion: N keys under one root in one proof.

Each statement is a guest binary that reuses the ldb reader (`Reader` in `crates/ldb/src/reader.rs`) and `apply` (`crates/ldb/src/apply.rs`) directly. The witnesses are the touched node chunks. Builds on the #301 foundations.

## Correction: what the substrate now provides

The walker is `nectar-ldb`, not `nectar-manifest`. #641 renamed the key-value crate to `nectar-ldb`, and the `nectar-manifest` name belongs to the format-agnostic trait seam carved by #644.

Three moves in the restructure change the crates this issue links, so read the target names from `docs/RESTRUCTURE-PLAN.md` rather than from the text above.

`nectar-proof` is created in M2 by epic #800 and supplies the authentication layer: how one step's bytes bind to a trusted address, plus the replay loop. That is the plumbing every statement here shares.

`nectar-ldb` splits in M2 into `nectar-ldb-core`, which is `no_std` and carries node decode, descent verification and the proof implementation, and `nectar-ldb` above it, which is the std engine. The reader and `apply` land on either side of that split, and #800 decides which. A guest links only the `no_std` half, so a transition statement that needs `apply` may need `apply` on the core side or a witness-driven re-derivation instead. Settle that before writing the transition guest.

The proving core is renamed: today's `nectar-primitives-core` (#614, landed by #642) becomes `nectar-primitives` in M2.

The `no_std` precondition holds. `nectar-ldb` is `#![no_std]` and the bare-metal `riscv64imac-unknown-none-elf` lane in `.github/workflows/nostd.yml` gates it. M0 narrows that matrix to the six proving-lane crates and no others, which are `nectar-primitives`, `nectar-spec`, `nectar-postage-primitives`, `nectar-proof` and `nectar-ldb-core`, plus `nectar-marker`.

mantaray is excluded from the proof seam permanently, because its roots are order-dependent for the same content. Every statement here is over ldb v1.

mfw78:
## What

Implements #302 as two compositional proofs built on the #301 inclusion/exclusion primitives, scoped entirely to the `nectar-manifest-proof` crate:

- **Range-completeness** (`src/range.rs`): `RangeProof`, `prove_range_complete(source, root, lo, hi)`, `verify_range<F>(root, lo, hi, proof)` - an authenticated ascending listing over `[lo, hi)`. The proof ships the frontier of trie nodes whose subtrees overlap the range, each self-authenticated by content address and chained from the trusted root. `verify_range` re-walks the frontier in key order, mirroring the ordered reader's flatten (value in its fork record, embedded children folded in place, referenced children as hops), and demands a witness node at every referenced child whose `[prefix, successor(prefix))` overlaps the range. A withheld or tampered node leaves an in-range edge with no witness, which is rejected via the new `VerifyError::NodeAbsent`, `VerifyError::Decode` and `VerifyError::Encrypted` variants.
- **State transitions** (`src/transition.rs`): `TransitionProof`, `Transition`, `prove_transition`/`prove_deletion`/`prove_update`, `verify_transition` - each transition is two single-key proofs read against different roots (`root_before`, `root_after`). An insertion is exclusion-then-inclusion, a deletion is its dual, and an update is inclusion under both roots with a changed value. Proving enforces the claimed shape (e.g. a false insertion where the key is already present, or still absent, cannot be constructed), and verification confirms both halves plus the shape.

`src/lib.rs` re-exports the new public surface and extends the module-level docs; `src/error.rs` gains the three `VerifyError` variants above. `tests/compositions.rs` adds coverage for both proof kinds. Neither `fuzz/` nor `nectar-postage-usage` is touched.

## Why

#301 gives single-key inclusion/exclusion proofs but no way to authenticate a whole range as complete (no witheld keys) or to attest what changed between two states. Both compositions reuse the existing primitives rather than adding new authentication machinery: range-completeness needs only content-address authentication plus the trie's total order (no additional cryptographic assumptions), and transition proofs are two ordinary single-key proofs paired under a shape constraint.

Closes #302

## Testing

Verified independently on branch `s264/92-proof-compositions` (HEAD `0bbd0fe`) atop base `s264/90-proof-core`, cloned fresh and checked out via `FETCH_HEAD`. No changes were needed; nothing was committed or pushed as part of verification.

Battery run via `nix develop --command`:
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean, no warnings.
- Tests: `cargo-nextest` was not present in the dev shell, so `cargo test --workspace --all-features` was substituted as a faithful pass/fail equivalent. All suites green.

A red-team pass on the frontier re-walk confirmed that every ancestor prefix of an in-range key overlaps `[lo, hi)`, so withholding any node on its path triggers `NodeAbsent`; overlap/successor half-open logic and the root-only entry guard are correct; tampering is caught via the re-sealed BMT address. No panics were found on adversarial `verify_range` input (empty, truncated, oversized, or flag-poking payloads all return errors).

This is a stacked PR targeting `s264/90-proof-core`, not `main`.

## AI Assistance

Claude (Anthropic, Sonnet 5) was used for implementation, red-teaming and independent verification of this change, including the battery run and the composition of this PR description.

### DIFF
--- a/crates/manifest-proof/src/error.rs
+++ b/crates/manifest-proof/src/error.rs
@@ -1,6 +1,6 @@
 //! Prove- and verify-side failures.
 
-use nectar_manifest::EncodeError;
+use nectar_manifest::{DecodeError, EncodeError};
 use nectar_primitives::{ChunkAddress, PrimitivesError};
 
 use crate::descent::DescentError;
@@ -64,4 +64,14 @@ pub enum VerifyError {
     /// encrypted subtree.
     #[error("proof path does not match the descent")]
     Malformed,
+    /// A frontier node the range walk needed was not carried in the proof; the
+    /// listing cannot be complete without it.
+    #[error("range node {0} absent from the proof")]
+    NodeAbsent(ChunkAddress),
+    /// A carried frontier node did not decode as a manifest node.
+    #[error(transparent)]
+    Decode(#[from] DecodeError),
+    /// The range spans an encrypted subtree the plain walk cannot enumerate.
+    #[error("range reaches an encrypted subtree")]
+    Encrypted,
 }
--- a/crates/manifest-proof/src/lib.rs
+++ b/crates/manifest-proof/src/lib.rs
@@ -22,6 +22,13 @@
 //! [`verify`] replays the descent over the authenticated bytes and reports what
 //! it finds, so a proof asserts no verdict its bytes do not.
 //!
+//! Two compositions ride the single-key primitives. A
+//! [range-completeness](prove_range_complete) proof attests a listing is every
+//! key in `[lo, hi)`, authenticating the frontier of nodes the range spans so an
+//! omitted key has no witness. A [state-transition](prove_transition) proof
+//! attests one key changed between two roots - an insertion, or its deletion and
+//! update duals - as its state under each root.
+//!
 //! The model is generic over [`Format`](nectar_manifest::Format) and defaults to
 //! the frozen `V1` plaintext layout.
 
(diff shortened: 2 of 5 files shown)


mfw78:

PR #311, opened 2026-07-13
feat(manifest-proof): add authenticated rank, count, select and pagination proofs

### LINKED ISSUE #303: feat(zkvm): order-statistic statements over the counted baseline in-guest
Deferred to Post v1 under epic #693. Rescoped 2026-07-18: the counted proof objects (PR #311) were closed in favour of zkVM guest proving, and the counted grammar is the baseline wire (`tag_version 0x01`), not a separate profile.

## What

Order-statistic statements in-guest over the counted baseline.

- `rank(key)`, `count(range)`, `select(i)` and `page(offset, limit)`, proven by descending with the on-wire subtree counts.
- The guest surfaces a count contradiction as a rejection, which is the `CountMismatch` posture, so a proof doubles as a count-consistency audit of the touched path.
- Optional whole-manifest count-consistency statement: one O(n) guest walk that certifies every count under a root.

Builds on the #301 foundations.

## The counted baseline in the tree

The counted grammar is the shipped `V1` parameter set: `VERSION = 0x01` in `crates/ldb/src/format.rs`, with the subtree count carried as `SubtreeCount` in `crates/ldb/src/count.rs`. The frozen parameters are pinned by test in the same file, so the wire the guest descends is fixed.

## Correction: the crates this issue links

`nectar-proof` is created in M2 by epic #800 and supplies the authentication layer, which is how one step's bytes bind to a trusted address, plus the replay loop. Order-statistic arithmetic is not in it by design: the plan states the crate carries no key vocabulary, no verdict and no span arithmetic. The rank and select arithmetic therefore belongs beside the counts, in `nectar-ldb-core`.

`nectar-ldb` splits in M2 into a `no_std` `nectar-ldb-core`, which carries node decode, descent verification and the proof implementation, and the std `nectar-ldb` engine above it. A guest links `nectar-ldb-core`. Read the target names from `docs/RESTRUCTURE-PLAN.md`.

mantaray is excluded from the proof seam permanently, because its roots are order-dependent for the same content. Every statement here is over ldb v1.

mfw78:
## What

Adds authenticated rank/count/select/pagination proofs over the baseline
counted wire grammar to `nectar-manifest-proof`. These proofs are generic
over `Format` (any manifest, since #307 makes every profile counted), not
scoped to a separate `0x03` profile as originally planned — that profile no
longer exists.

New module `crates/manifest-proof/src/counted.rs` (~1000 lines) provides:

- `prove_rank` / `verify_rank`
- `prove_count` / `verify_count`
- `prove_select` / `verify_select`
- `prove_page` / `verify_page` and `prove_page_prefix` / `verify_page_prefix`

with public types `RankProof`, `CountProof`, `SelectProof`, `PageProof` and
`CountedPath`, all re-exported from `lib.rs`.

A proof is the O(depth) chain of node payloads the count-descent visits, each
content-addressed and chained from the trusted root. Verification re-derives
the answer from the authenticated bytes alone, never trusting a claim carried
in the proof: the flatten parser matches the manifest codec's `ForkRecord`
wire order and reads the trailing `child_count` / `seg_count` via
`wire::Cursor`. `count` is computed as `rank(hi) - rank(lo)` (two rank
descents); `prove_page` carries both bound rank descents plus one rank-pinned
select descent per returned key, so the verifier pins every returned entry to
the counted commitment.

`descent.rs` gains the count-aware descent support these proofs walk, and
`error.rs` adds the new failure modes: `ProveError::IndexOutOfRange`, and
`VerifyError::CountMismatch`, `VerifyError::IndexOutOfRange`,
`VerifyError::PageShape`.

## Why

Closes #303

The count-independent inclusion/exclusion/range-completeness proofs already in
this crate cannot answer order-statistic questions (rank, count, select,
paginated listing) without an O(n) scan. Since the baseline wire now commits
subtree counts at every fork (#307), an O(depth) authenticated descent can
answer these directly over any manifest, not only a separate profile. As
documented in `lib.rs`, this family trusts an HONEST BUILDER: a referenced
child's count is author-asserted and bound to its parent chunk, but not
checked against the child's real subtree, so these proofs establish the count
as committed by the root rather than against an adversarial root. Every hop
the descent does fetch cross-checks the parent-asserted `child_count` against
the child's own derived total and rejects an on-path inconsistency; only the
counts of un-fetched siblings a descent skips remain purely author-asserted.
The strictly trustless answers continue to ride the existing count-independent
proofs.

## Testing

New `tests/counted.rs` (513 lines) covers the rank/count/select/page proof
and verify paths, alongside the existing `tests/compositions.rs` and
`tests/roundtrip.rs` suites.

Full battery run via `nix develop`:

- `cargo clippy --workspace --all-targets --all-features -- -D warnings`:
  clean.
- `cargo nextest run -p nectar-manifest -p nectar-manifest-proof --locked`:
  234 tests run, 234 passed, 0 skipped.
- `cargo test --doc --workspace --locked`: clean.
- A stale-reference sweep at the tip (`counted grammar's trailing`, `counted
  profile`, `tag_version 0x03`) returns zero hits.

## AI Assistance

Implemented with Claude Sonnet 4.5 and adversarially reviewed with Claude Opus 4.5, per the process in [nxm-rs/.github CONTRIBUTING.md](https://github.com/nxm-rs/.github/blob/main/CONTRIBUTING.md). The red-team pass found no real bugs: the flatten parser was checked byte-for-byte against the manifest codec's `ForkRecord` wire order, adversarial input is `Cursor`-gated with no indexing, `unwrap` or `panic`, and the rank/count/select/page algorithms were checked against the independent reader oracle in `order.rs`.

### DIFF
--- a/crates/manifest-proof/src/counted.rs
+++ b/crates/manifest-proof/src/counted.rs
@@ -1,0 +1,998 @@
+//! Authenticated rank, count, select and pagination proofs over the counted
+//! node grammar.
+//!
+//! The node grammar carries every subtree's distinct key-count in the
+//! authenticated node bytes (spec 5.6), so an order-statistic answer is provable
+//! with the same O(depth) descent the reader walks: a [rank](prove_rank) or
+//! [select](prove_select) proof is the chain of nodes the descent visits, each
+//! authenticated against the reference the one before it yielded, and the
+//! verifier re-derives the answer from those bytes alone. A [count](prove_count)
+//! proof is two rank descents whose difference is the window size; a
+//! [page](prove_page) proof pins a listing slice by its absolute ranks.
+//!
+//! TRUST BOUNDARY - counted proofs assume an HONEST BUILDER. A referenced
+//! child's `child_count` is bound to its parent chunk's bytes but NOT to the
+//! child's real subtree: re-encoding a chunk reproduces the same count, it does
+//! not re-derive it from the child, and an inflated count stays internally
+//! consistent all the way to the root. So these proofs establish the count as
+//! COMMITTED BY THE ROOT - sound against an honest builder (a gateway, indexer
+//! or registry operator whose manifest you already rely on), NOT against an
+//! adversarial root that plants false counts on the subtrees a descent skips.
+//! The strictly trustless answers ride the count-independent inclusion,
+//! exclusion and range-completeness proofs, which are sound against adversarial
+//! roots by canonicalness.
+//!
+//! The residual trust is bounded where free: every hop the descent DOES take
+//! fetches the child, so the verifier cross-checks the parent-asserted
+//! `child_count` against the child's own derived total (the sum of its fork
+//! counts) and rejects an on-path inconsistency. Only the counts of the
+//! un-fetched siblings a descent skips remain purely author-asserted.
+
+use nectar_manifest::{Entry, Format, Key, SubtreeCount};
+use nectar_primitives::wire::Cursor;
+use nectar_primitives::{
+    ChunkAddress, ChunkOps, ChunkRef, ContentChunk, DEFAULT_BODY_SIZE, EncryptedChunkRef,
+};
+
+use crate::descent::{DescentError, Width, child_width, entry_width, flag, take_u16, take_value};
+use crate::error::{ProveError, VerifyError};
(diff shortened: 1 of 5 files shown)


mfw78:

PR #312, opened 2026-07-13
feat(contracts): add Solidity MantarayProofVerifier for on-chain proofs

### LINKED ISSUE #304: feat(zkvm): on-chain verification through the sp1 and risc0 standard verifiers
Deferred to Post v1 under epic #693. Rescoped 2026-07-18: the bespoke Solidity verifier (PRs #312 and #313) was closed, and on-chain verification goes through the zkVMs' standard verifiers instead.

## What

On-chain verification of guest-proven manifest statements.

- Wrap guest proofs to Groth16 or PLONK, and verify through the SP1 verifier gateway and the risc0-ethereum verifier.
- Design the journal and public-values ABI for each statement family, which are inclusion, exclusion, range, transition and count. Each carries the root, a statement discriminant and a statement payload.
- A thin statement-decoding contract per family over the shared verifier, which replaces per-statement verifier logic.
- Gas measured per statement family.

## Notes on the tree as it stands

There is no Solidity in this repository today. `nectar-contracts` is Rust bindings only, so the contracts here are new work rather than an edit of something shipped.

`nectar-contracts` is std and stays std under the restructure. It sits at tier 3, depends on `nectar-spec`, and it is not on the six-crate proving lane. A guest therefore cannot link it, and any chain state a statement needs must cross into the guest as plain witness data.

Depends on #301 for the guest programs, and on #302 and #303 for the aggregate and order-statistic families.

mfw78:
## What

Adds a reusable Solidity `MantarayProofVerifier` library (new Foundry project at repo-root `contracts/`) that verifies mantaray 1.0 inclusion and exclusion proofs on-chain, at BMT-segment granularity, against a trusted root.

- `contracts/src/MantarayProofVerifier.sol` exposes `verifyInclusion(root, key, value, proof) -> bool`, `verifyExclusion(root, key, proof) -> bool`, and `decode(bytes) -> Proof`. It follows the key itself and accepts a node only when its address equals the reference the previous authenticated hop supplied, so a whole proof is a BMT hash chain anchored at the root.
- The library targets the frozen `tag_version 0x01` plaintext layout only:
  plaintext 32-byte references (`ref32`), and non-spilled nodes (flag bits
  5/6 rejected). Subtree counts on the wire are skipped, not verified: the
  descent bounds records by fork-index offsets, so count bytes are crossed
  but never read. `contracts/README.md` documents the restricted profile and
  what encrypted subtrees, the `0x02` read profile, on-chain counted
  (rank/select/count/page) answers, and spilled segment directories would
  each need to be supported.
- `_bmtRoot` mirrors nectar's prefix-free content-chunk BMT: keccak256 up the binary tree of 128 32-byte segments, then `keccak(span_le[8] || tree_root)`.
- `contracts/foundry.toml`, `.gitignore`, `.gas-snapshot` and a `test/` suite (`MantarayProofVerifier.t.sol`, `Harness.sol`) are included. Solidity is supplied by nix (`FOUNDRY_SOLC=$(which solc)` under `nix-shell -p foundry solc`); no `svm` download and no `forge install` — `test/Harness.sol` vendors a minimal cheatcode interface and assertion helpers instead of pulling in forge-std.
- `crates/manifest-proof/tests/solidity_fixtures.rs` adds an `emit_solidity_fixtures` test that (re)generates the binary fixtures under `contracts/test/fixtures/` from the Rust proof generator, so the wire format is single-sourced from the Rust side.

## Why

On-chain consumers need to verify mantaray inclusion/exclusion proofs without re-implementing the Rust descent logic from scratch or trusting an off-chain relayer. This library ports the `crates/manifest-proof` descent (`descent.rs`/`verify.rs`) and wire codec (`codec.rs`) to Solidity so a contract can authenticate a proof directly against a root address.

## Testing

- `contracts/test/MantarayProofVerifier.t.sol` covers inclusion (embedded and referenced), the three exclusion terminal conditions (gap, divergent, exhausted) plus an empty-key and a referenced exclusion case, and negative cases (tampered proof, wrong root, wrong value) — ten tests total, gas figures recorded in `contracts/.gas-snapshot`.
- Fixtures are generated by `cargo test -p nectar-manifest-proof --test solidity_fixtures` and consumed directly by the Foundry suite via `readFileBinary`, keeping the Rust and Solidity sides of the wire format in lockstep.

This PR is stacked on `s264/94-proof-counted` and targets that branch, not `main`.

Closes #304

AI Assistance: Claude (Anthropic) used for implementation of the Solidity verifier library, its Foundry test suite, the Rust fixture emitter, and this PR description.


mfw78:

PR #313, opened 2026-07-13
feat(contracts): add demo verifier contracts

### LINKED ISSUE #305: feat(zkvm): end-to-end prove-and-verify demo pipeline
Deferred to Post v1 under epic #693. Rescoped 2026-07-18: the demo verifier contracts (PR #313) were closed with the bespoke Solidity verifier, and the demo target is the zkVM pipeline end to end.

## What

An end-to-end demonstration: build a manifest, prove a statement in the guest under both SP1 and RISC Zero, wrap it, and verify it on-chain against anvil. Cover at least inclusion and one aggregate statement, which is range or transition.

Record cycle counts, proof sizes, wrapping time and verification gas in the run output.

## Sequencing

Depends on #301 for the guest foundations, on #302 or #303 for the aggregate statement, and on #304 for the on-chain verifier path.

The manifest under test is an ldb v1 database. mantaray is excluded from the proof seam permanently, because its roots are order-dependent for the same content.

The measured output of this run is the input to #306, so emit it as JSON rather than as prose.

mfw78:
Closes #305

## What

Four thin demo contracts over the #304 `MantarayProofVerifier` library, each parsing its arguments, calling the verifier, and applying the effect, with no business logic of its own:

- `Registrar.sol` — claims a name only against an exclusion proof it is absent under the current root, then advances the root (a squatting-proof registrar).
- `RevocationRegistry.sol` — `isValid` gates on an exclusion proof a serial is absent from the revocation manifest (a trustless CRL/OCSP); `isRevoked` is the inclusion-gated dual.
- `Listing.sol` — accepts a directory listing only against a range-completeness proof, recording the verifier-derived digest keyed by the listing's bounds.
- `Log.sol` — appends a key only against a state-transition proof (absent under the prior head, present under the new one), giving an append-only log that cannot equivocate.

`Listing.sol` needs a new companion verifier, `MantarayRangeVerifier.sol`: a chunk-granularity range-completeness verifier. It re-BMTs whole frontier node payloads (prefix-free 128-leaf content-chunk BMT, `keccak(spanLE||root)`), matches each carried node by its derived content address, and re-walks the frontier (root value, fork table in key order, embedded children folded in place, referenced children demanded when their subtree overlaps `[lo, hi)`).

`MantarayProofVerifier.sol` gains `verifyTransition(rootBefore, rootAfter, key, value, before, after)`, composing an exclusion proof at `rootBefore` with an inclusion proof at `rootAfter` over the existing single-key segment machinery, which is what `Log.sol` rides.

## Why

#305 asks for demo contracts that first-class the accepted use cases of the verifier library from #304, so the exclusion, inclusion, state-transition and range-completeness primitives each have a minimal, legible on-chain consumer rather than only test coverage.

## Testing

- `forge test` over the new suites: `Registrar.t.sol`, `RevocationRegistry.t.sol`, `Log.t.sol`, `Listing.t.sol`, and `DemoTest.sol`, covering both the happy path and rejection (a taken name, an incomplete listing, a false transition) for each contract.
- `contracts/test/fixtures/range_all.bin`, `range_prefix.bin` and `transition_insert.bin` are new fixtures emitted by the same Rust `nectar-manifest-proof` generator as the existing single-key fixtures, exercised via `crates/manifest-proof/tests/solidity_fixtures.rs`.
- `contracts/.gas-snapshot` updated for the new call paths.

Stacked on #s264/96-solidity-verifier; no CI beyond CLA runs on this branch until it is retargeted onto that base.

AI Assistance: Claude used for implementation, red-team review, and testing verification.


mfw78:

PR #316, opened 2026-07-17
ci: gate the no_std core on a bare-metal rv64 target

mfw78:
## What

Adds a `no_std` workflow: `cargo check --no-default-features --target riscv64imac-unknown-none-elf` for the crates that already build bare-metal (`nectar-swarms`, `nectar-contracts`). The `-p` list grows as the remaining crates flip to no_std.

## Why

nectar's guiding decision is full library-core no_std+alloc (alloy/revm-style) targeting zkVM guests, and the rv64 bare-metal shape matches where the guest ecosystem is heading. Nothing in CI built `--no-default-features` before, so the declared no_std surfaces rot silently; this stops that for the crates that are already there, and gives the post-train primitives no_std epic a gate to grow into.

## Testing

- Verified locally on the pinned 1.94 toolchain: both crates check clean for `riscv64imac-unknown-none-elf` in under 10 s.
- `actionlint` clean.
- The job exercises itself on this PR's own CI run (warm-cache cost is roughly a minute, in parallel with existing jobs).

## AI Assistance

Drafted with AI assistance; design, review, and testing by the author.

### DIFF
--- a/.github/workflows/nostd.yml
+++ b/.github/workflows/nostd.yml
@@ -1,0 +1,34 @@
+# Gates the no_std core on a bare-metal rv64 target (the zkVM-guest shape).
+# Only the crates that already build without std are listed; extend the -p
+# list as crates flip to no_std.
+name: no_std
+
+on:
+    pull_request:
+    merge_group:
+    push:
+        branches: [main]
+
+env:
+    CARGO_TERM_COLOR: always
+
+concurrency:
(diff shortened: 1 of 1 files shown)


mfw78:
Superseded by current main: the capability this proposes already exists there (the riscv64imac no_std lane in nostd.yml / the cargo clippy --workspace --all-targets --locked pass in lint.yml), delivered through the later CI infrastructure work, and this branch is now conflicting. Closing as superseded.

PR #321, opened 2026-07-18
ci: lint every target with clippy

mfw78:
## What

Adds `--all-targets` to the clippy job, which previously linted only libs and bins: every test, bench and example in the workspace escaped all lints. Fixes the one unused import the wider net surfaces (`Chunk` in `mantaray/builder.rs`). Three unused-macro warnings in the feature-gated joiner/splitter test helpers remain visible deliberately: the async-testing rework converts their only call sites to ungated tests, making them used again.

## Why

The deny-level restriction lints and the upcoming `disallowed-methods` gate for `block_on` are meaningless over test code that clippy never sees. Warnings do not gate (no `-D warnings` change here), so this widens coverage without altering merge semantics.

## Testing

`cargo clippy --workspace --all-targets --locked` exits clean locally (warnings only, no deny-lint failures); 119 mantaray tests pass after the import fix; actionlint clean.

## AI Assistance

Written with AI assistance; reviewed and tested by the author.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,11 +25,12 @@ jobs:
             - uses: ./.github/actions/rust-setup
               with:
                   components: clippy
-            # Lints lib+bins with default features; the `deny`-level restriction
-            # lints in `[workspace.lints.clippy]` make this fail on any
-            # panicking/overflowing operation in production code.
+            # Lints every target (libs, bins, tests, benches, examples) with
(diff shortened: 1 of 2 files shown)


mfw78:
Superseded by current main: the capability this proposes already exists there (the riscv64imac no_std lane in nostd.yml / the cargo clippy --workspace --all-targets --locked pass in lint.yml), delivered through the later CI infrastructure work, and this branch is now conflicting. Closing as superseded.

Issue #323, opened 2026-07-18
epic: streaming file pipeline (nectar-file)


mfw78:
Extracts a new `#![no_std]` + alloc `nectar-file` crate built on two dataflow machines: a bounded fetch window walking down (frame stack, sequence-tagged), and a spill-on-close level frontier writing up (bounded put window). All public read/write modes are drain policies over these two engines, not separate implementations.

The core stays poll-native and io-free; tokio/rayon/std adapters wrap the poll surface without introducing spawns, channels or timers into core. rv64imac and wasm32 no_std lanes are required from the first sub-issue.

This epic opens only after its foundational CI issue (rv64imac/wasm32 no_std dependency spike) is green; it sits on the standalone precursor #322 for the `Chunk<Trust, B>`/`TrustedGet` seam.

mfw78:
All 21 sub-issues complete: the streaming file pipeline (nectar-file) landed end to end — geometry, walk/split engines, file facade, tokio/rayon adapters, sinks, encrypted mode, manifest bridge, benches, exhibits, fuzz battery, and the legacy-surface removal (#389, closing #351). The `Split::collect` family replaced the deleted legacy one-shot splitter.

PR #361, opened 2026-07-18
test: convert primitives/mantaray/manifest to async house pattern

### LINKED ISSUE #104: chore(primitives): silence unused-macro clippy warnings under default features
## Summary

`cargo clippy --workspace --all-targets -- -D warnings` fails under the default feature set on two unused-macro warnings in `nectar-primitives`:

- `unused macro definition: generate_encrypted_joiner_tests` at `crates/primitives/src/file/joiner_tests.rs:150`
- `unused macro definition: generate_encrypted_splitter_tests` at `crates/primitives/src/file/splitter_tests.rs:79`

## Detail

Both macros are only used under the `encryption` feature. Without that feature their sole call sites are compiled out, so the macro definitions appear unused and clippy fails the `--all-targets` build with `-D warnings`.

This is pre-existing on `main` and is unrelated to PR #103 (the postage sync-core change). PR #103 only surfaces it in passing: the warning appears whenever `--all-targets` is built without the `encryption` feature, independently of that work.

## Proposed action

Silence the warnings for non-`encryption` builds, for example by `cfg`-gating the macro definitions on the `encryption` feature, or by annotating them with `#[allow(unused_macros)]` (or `#[cfg_attr(not(feature = "encryption"), allow(unused_macros))]`) so a default-feature `-D warnings` build stays clean while the macros remain available under `encryption`.

### LINKED ISSUE #353: test: route test executor entry through nectar_testing::run
Route every test executor entry point through `nectar_testing::run`, so the workspace has one sanctioned choke point and #362's `disallowed-methods` gate can land.

This is a mechanical rename, not a body conversion. `crates/testing/src/lib.rs` defines:

```rust
#[allow(clippy::disallowed_methods)]
pub fn run<F: Future>(f: F) -> F::Output { futures_executor::block_on(f) }
```

which is a drop-in shim with a signature identical to `futures::executor::block_on`. Satisfying the gate therefore costs a token rename, not a rewrite of test bodies.

## Scope

- `block_on($X)` becomes `run($X)`, in place. Roughly 468 sites: an `ast-grep` pass gets about 366, and the rest sit inside macro bodies (`assert!`, `assert_eq!`, `proptest!`, `fuzz_target!`) which tree-sitter treats as opaque token trees, so a regex pass finishes them.
- `use futures::executor::block_on` becomes `use nectar_testing::run`, with `nectar-testing` added to the affected crates' dev-dependencies.
- Two genuine collisions, the only ones in the workspace: a local `let run = vec![..]` binding in `crates/manifest/src/apply.rs` shadows the function, and `rt.block_on` in `crates/file/benches/latency_bench.rs` must stay as it is, being a deliberate tokio bench runtime.
- Delete the five byte-identical local `yield_now` helpers (`file/src/split/tests.rs`, `file/src/walk/tests.rs`, `file/src/parallel/tests.rs`, `file/src/read/cancel.rs`, `file/tests/membound.rs`) in favour of `nectar_testing::yield_now`. Note the lint cannot see a local `fn`, so this is not gate work, just genuine duplication.

After the rename the gate leaves six violations, all `tokio::runtime::Runtime::block_on`: three in the tokio bench, which take an `#[allow]` here, and three from `#[tokio::test]` expansion in `postage-issuer`, which #362 already allows.

## Why not the body conversion

The original scope proposed converting 418 sites to `($X).await` inside `async fn` bodies wrapped in `run`. Four review passes rejected that:

- **It changes nothing.** `run` is `block_on`, so the conversion is behaviour-preserving by construction. `nectar-testing` currently has no dependents, and no hang, flake or bug in the history would have been prevented.
- **It is unsafe in specific places.** `crates/file/src/read/cancel.rs` builds its contract on manual polling with a noop waker, and passes a *sync* closure into a helper. Wrapping those bodies in `run` nests `block_on` inside `block_on` on one thread; `futures-executor` shares a thread-local `ThreadNotify`, so the inner loop consumes the outer's wakeup token and the test parks forever. `run` has no budget, so that is a silent CI timeout rather than a red test. The `park_forever` tests are safe only because they are polled a bounded number of times with a noop waker.
- **It cannot land.** #361 lost 85% of its file surface in 4.04 days; main gains roughly 140 new `block_on` sites per day, and the train's own zero-conflict PR still took 3.72 days to merge. A rename is regenerable by re-running the script; a 10,000-line body conversion is not.
- **Review would be nominal.** #361 ran to +5524/-4277 for about 250 sites, roughly 20 changed lines each, restructuring test bodies. A diff that size labelled mechanical gets approved on sight, trading real coverage for a green lint.

The body conversion is still worth doing where it changes what a test exercises, in particular the roughly 135 sites inside loops. That is split out and carries no gate dependency.

## Related

- Fix the `nectar-testing` module doc: it claims "a deadlock becomes a failure instead of a hang", which is true of `yield_until` and false of `run`. Consider giving `run` a poll budget that panics, which would make the claim true and turn the hazards above into loud failures.
- #361 is closed rather than rebased: 12 of its 43 files no longer exist and 22 were rewritten.
- `warn` mode is not an option: `lint.yml` runs clippy without `-D warnings`, so a warn-level gate never fails CI.

Blocks: #362
Depends on: #352
Size: S

mfw78:
## What

Converts the primitives, mantaray, and manifest crates to the async house pattern in three per-crate commits (`af83aefd`, `c575daff`, `ffb4df94`), plus a follow-up fix commit (`ed390d3a`):

- Replaced `block_on` litanies in tests with `nectar_testing::run(async { .. })`, awaiting calls inline.
- Moved proptest closures to the sync-over-sim shape: `run()` inside `proptest!`, with an async block returning `Ok::<(), TestCaseError>(())` and `?` propagation.
- Switched criterion benches (`file_bench`, `encryption_bench`, `mantaray_bench`) to `to_async(FuturesExecutor)` for timed sections and `nectar_testing::run` for setup/corpus building; added the `async_futures` feature to the workspace criterion dependency.
- Split `joiner.rs`'s `cfg(all(test, feature = "tokio"))` module: runtime-agnostic probes moved into a plain `#[cfg(test)] mod tests` (with `tokio::task::yield_now` replaced by `nectar_testing::yield_now`, and `current_thread` flavors dropped).
- Fixed a half-converted rustdoc example in `manifest/src/dx.rs` that imported `nectar_testing::run` but kept a `block_on` body, which broke `cargo test --doc -p nectar-manifest`.

## Why

Brings primitives, mantaray, and manifest onto the async house pattern established elsewhere in the workspace, removing runtime-specific (`block_on`/tokio-only) test scaffolding in favour of the shared `nectar_testing` harness.

Closes #353

## Testing

- `cargo clippy --workspace --all-targets --locked`: two pre-existing warnings only (`use_self` in `primitives/src/chunk/type_tag.rs:110`, encryption-gated unused import in `mantaray/src/builder.rs:45`), both verified present on the base branch (`origin/s325/12-wasm-smoke`) and byte-identical to base in this diff.
- `cargo nextest run -p nectar-manifest -p nectar-mantaray -p nectar-primitives --all-features`.
- `cargo test --doc -p nectar-manifest` (regression check for the `dx.rs` fix).

## AI Assistance

Implementation by Fable, review by Opus, PR by Sonnet.

Closes #104

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -68,11 +68,37 @@ jobs:
                     -p nectar-manifest --locked \
                     --target wasm32-unknown-unknown
 
+    wasm-smoke:
+        # Executable proof of the wasm store-trait relaxation: a !Send store
+        # satisfies ChunkGet only on wasm32, so the smoke test must actually run
+        # there. Executes under Node via wasm-bindgen-test-runner, which nextest
+        # cannot host; hence a separate plain-cargo job. The runner version must
+        # equal the locked wasm-bindgen version (kept in lockstep by the
+        # workspace's exact wasm-bindgen-test pin).
+        name: wasm / node smoke
+        runs-on: ubuntu-latest
+        timeout-minutes: 30
+        steps:
+            - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+            - uses: ./.github/actions/rust-setup
+              with:
+                  targets: wasm32-unknown-unknown
+            - uses: taiki-e/install-action@43aecc8d72668fbcfe75c31400bc4f890f1c5853  # v2.83.2
+              with:
+                  tool: wasm-bindgen@0.2.122
+            - name: Run wasm smoke tests
+              env:
+                  CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER: wasm-bindgen-test-runner
+              run: |
+                  cargo test \
+                    -p nectar-testing --test wasm_smoke --locked \
+                    --target wasm32-unknown-unknown
+
     unit-success:
         name: unit success
         runs-on: ubuntu-latest
         if: always()
-        needs: [test, wasm]
+        needs: [test, wasm, wasm-smoke]
         timeout-minutes: 30
         steps:
             - name: Decide whether the needed jobs succeeded or failed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2049,6 +2049,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6b947ae49db0d222b1dbc6b113ce7248a3fc3a6ca21b696717bfc000ba4484d8"
 
 [[package]]
+name = "minicov"
+version = "0.3.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4869b6a491569605d66d3952bcdf03df789e5b536e5f0cf7758a7f08a55ae24d"
+dependencies = [
+ "cc",
+ "walkdir",
+]
+
+[[package]]
 name = "nectar-contracts"
 version = "0.4.0"
 dependencies = [
@@ -2193,7 +2203,19 @@ dependencies = [
 name = "nectar-testing"
 version = "0.4.0"
 dependencies = [
+ "futures",
  "futures-executor",
+ "nectar-primitives",
+ "wasm-bindgen-test",
+]
+
+[[package]]
+name = "nu-ansi-term"
+version = "0.50.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
+dependencies = [
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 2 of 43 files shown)


mfw78:
Parked as a draft and superseded. This was written against a pre-#323 tree: 12 of its 43 files no longer exist on `main` (the legacy `primitives/src/file/` pipeline and the legacy mantaray `builder.rs`/`manifest.rs` facade), and it does not touch `nectar-file`, which the #323 train built afterwards and which now carries 87 of the 418 violations. Measured against the gate #362 introduces, it converts 23 of 47 affected files, so landing it would leave #362 failing or needing blanket allowances. #353 has been rescoped to the measured violation set and the sweep will be redone against current `main`.

mfw78:
Closing rather than rebasing. Of its 43 files, 12 no longer exist on `main` (the legacy `primitives/src/file/` pipeline and the mantaray `builder.rs`/`manifest.rs` facade) and 22 have been rewritten, so a rebase would be a from-scratch rewrite wearing a rebase costume.

Superseded by a different approach. `nectar_testing::run` is already a drop-in shim with a signature identical to `futures::executor::block_on`, so the #362 gate can be satisfied by a mechanical rename rather than a 418-site body conversion: 468 renames, 2 hand fixes, net zero lines, workspace tests green, and the gate clears with one `#[allow]` in a bench. The body-conversion work is genuinely valuable in places but is test-quality work to be scoped per test, not a gate prerequisite, so it is being split out. See #353.

PR #362, opened 2026-07-18
chore: gate ad-hoc block_on/yield_now via clippy disallowed-methods

### LINKED ISSUE #354: test: enforce the block_on gate in clippy
Adds `clippy.toml` disallowed-methods covering `futures_executor::block_on` and the `futures::executor::block_on` re-export (canonical def paths, so the entries actually resolve), plus `tokio::runtime::Runtime::block_on`/`Handle::block_on` (exempting the two `perf_v3.rs` paused-clock sites), plus `tokio::task::yield_now`. Adds a `--features tokio` clippy pass for the primitives/postage adapter modules. Verify the lint fires on a known site before merging.

Builds on the already-landed `--all-targets` lint.yml change, which is what makes the disallowed-methods lint see test/bench code at all.

Depends on: #352, #353
Size: S

mfw78:
## What

- Add `clippy.toml` with `disallowed-methods` covering `futures::executor::block_on`/`futures_executor::block_on` (both re-export and canonical paths), `block_on_stream` (both paths), `LocalPool::run_until` (both paths), `tokio::runtime::Runtime::block_on`, `tokio::runtime::Handle::block_on`, and `tokio::task::yield_now`.
- Set `disallowed_methods = "deny"` in `[workspace.lints.clippy]` (CI clippy runs without `-D warnings`, so `warn` would never gate).
- `lint.yml`: upgrade the base pass to `--all-targets`, and add a second pass with `--features nectar-primitives/tokio,nectar-postage-usage/client` so the non-default tokio adapter modules are linted too.
- Add sanctioned `#[allow(...)]` at 5 call sites where the `#[tokio::test]`/`#[tokio::main]` macros expand to `Runtime::block_on` (`crates/postage-issuer/src/factory.rs`, `crates/postage-usage/examples/roam_between_machines.rs`, `crates/postage-usage/src/client.rs`, `crates/primitives/src/file/joiner.rs`, `crates/primitives/src/file/windowed.rs`).

## Why

Ad-hoc executor entry points (`block_on`, `LocalPool::run_until`, blocking `yield_now`) bypass the deterministic manual-poll test harness (`nectar_testing::run`), which can hide non-determinism and cancellation bugs. This gate makes any new such call site fail CI.

Closes #354

## Testing

- `cargo clippy --workspace --all-targets --locked`: exit 0. Only two pre-existing warn-level warnings surface (`use_self` in `primitives/src/chunk/type_tag.rs:110`, unused import in `mantaray/src/builder.rs:45`), both in files untouched by this branch.
- `cargo clippy --workspace --all-targets --locked --features nectar-primitives/tokio,nectar-postage-usage/client`: exit 0.
- Gate proven live: removing one of the sanctioned `#[allow]` sites causes clippy to emit `error: use of a disallowed method \`tokio::runtime::Runtime::block_on\`` (3 deny errors), confirming the def-paths resolve and the allows are load-bearing.

## AI Assistance

Implementation by Fable, review by Opus, PR description by Sonnet.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,23 +25,29 @@ jobs:
             - uses: ./.github/actions/rust-setup
               with:
                   components: clippy
-            # Lints lib+bins with default features; the `deny`-level restriction
-            # lints in `[workspace.lints.clippy]` make this fail on any
-            # panicking/overflowing operation in production code.
+            # Lints every target (libs, bins, tests, benches, examples) with
+            # default features; the `deny`-level restriction lints in
+            # `[workspace.lints.clippy]` fail on panicking/overflowing
+            # operations, with test code exempted per crate.
             - name: cargo clippy
-              run: cargo clippy --workspace --locked
+              run: cargo clippy --workspace --all-targets --locked
             # The async io adapters sit behind the tokio feature, outside the
             # default-features pass above.
             - name: cargo clippy (tokio adapters)
-              run: cargo clippy --locked -p nectar-file --features tokio
+              run: cargo clippy --locked --all-targets -p nectar-file --features tokio
             # The batch ingest sits behind the rayon feature, likewise
             # outside the default-features pass.
             - name: cargo clippy (rayon ingest)
-              run: cargo clippy --locked -p nectar-file --features rayon
+              run: cargo clippy --locked --all-targets -p nectar-file --features rayon
             # The encrypted split mode sits behind the encryption feature,
             # likewise outside the default-features pass.
             - name: cargo clippy (encrypted split)
-              run: cargo clippy --locked -p nectar-file --features encryption
+              run: cargo clippy --locked --all-targets -p nectar-file --features encryption
+            # The postage-usage tokio surfaces are feature-gated, so their tests
+            # are invisible to the passes above; compile them in so the
+            # disallowed-methods gate sees them.
+            - name: cargo clippy (postage-usage surfaces)
+              run: cargo clippy --workspace --all-targets --locked --features nectar-postage-usage/client,nectar-postage-usage/issuer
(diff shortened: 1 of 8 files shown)


mfw78:
Rebased onto #423 (`test/adopt-nectar-testing`) and brought up to current `main`.

Two of the five sanctioned allow sites, `primitives/src/file/joiner.rs` and `primitives/src/file/windowed.rs`, no longer exist: the #323 train removed the legacy file pipeline, so those hunks are dropped. `nectar-primitives` also no longer has a `tokio` feature, so the second clippy pass now names only `nectar-postage-usage/client,issuer`. Both CI workflows were unioned with main's current nectar-file feature passes rather than replacing them, and `--all-targets` is added throughout so the gate sees test code.

Verified on the stack: `cargo clippy --workspace --all-targets --locked` and the postage-usage feature pass both report **zero** disallowed-method violations at `deny`. The gate is green because #423 lands the rename first, which is why this is stacked rather than based on main.

mfw78:
Rippled onto `main` after #423 merged, and the gate found a real hole in this PR.

The two allow sites this PR carried for `primitives/src/file/{joiner,windowed}.rs` were the tokio adapter tests. The #323 train moved that surface to `crates/file/src/tokio/tests.rs`, which had no allow, so with `--all-targets` on the tokio feature pass the gate reported 11 `Runtime::block_on` violations there. Added the sanctioned allow on the `mod tests` declaration, matching the phrasing used for the other adapter sites.

Worth noting the hole only appeared because this PR now adds `--all-targets` to the nectar-file feature passes when resolving against main's current workflow. Without it those adapter tests would have been invisible to the gate entirely, so the gate would have shipped green while not actually covering them.

Verified on `main`: all five clippy passes (default, postage-usage features, and nectar-file tokio/rayon/encryption) report zero disallowed-method violations at `deny`.

PR #390, opened 2026-07-19
fix(manifest): recanonicalize a re-capped chain tail so apply matches rebuild

### LINKED ISSUE #320: fix(manifest): apply diverges from a from-scratch rebuild on long chained keys after a remove
Proptest counterexample (found during unrelated gate runs, bisected to reproduce identically at current main and one commit earlier, so latent rather than train-caused): with base key set `{[0,0]}`, applying a changeset that removes `[0,0]` and inserts a 253-byte chained key yields an `apply` root that differs from the from-scratch rebuild root for the merged key set. This violates the history-independence contract on apply's long-key/chained-edge path.

Replay: append this line to `crates/manifest/tests/apply.proptest-regressions` and run `cargo nextest run -p nectar-manifest -E 'test(apply_equals_rebuild_over_long_keys)'`:

```
cc f4a05b32307a56b322ea29e18cfd8396749df177a140cf39d9236e70c53da374
```

The regression line should be committed together with the fix so the case stays pinned.

mfw78:
## What

Recompact the tail emitted by chain()'s base case in apply.rs. When compact() merges an outer edge with a chain-head record built at a deeper boundary, the re-capped PLEN_MAX boundary can land earlier than the boundary the record's embedded child was built at; the base case previously emitted the stale payload verbatim, leaving a one-byte chain hop that a from-scratch build merges into a single edge. The recompaction recurses through any number of stale boundary levels, so the applied root is byte-identical to the rebuilt root.

## Why

Manifests are history-independent by contract: applying a changeset to a persisted root must produce the same root as rebuilding from the merged key set. The divergence was found by the apply-vs-rebuild property (proptest seed committed) and reproduced by the fuzz battery; long chained keys crossing the prefix cap after a remove-and-insert pair triggered it.

Closes #320

## Testing

- New fixed-vector unit test a_remove_beside_a_recapped_chain_insert_equals_a_rebuild (257-byte chained key beside a removed sibling).
- The failing proptest seed cc f4a05b32... is committed to apply.proptest-regressions and replays green through apply_equals_rebuild_over_long_keys.
- cargo nextest -p nectar-manifest: 211 passed; clippy --all-targets and fmt clean.

## AI Assistance

Implementation and diagnosis Fable, conducted review; filed by the session conductor.

### DIFF
--- a/crates/manifest/src/apply.rs
+++ b/crates/manifest/src/apply.rs
@@ -577,6 +577,15 @@ where
     F: Format,
 {
     if prefix.len() <= F::PLEN_MAX {
+        // Re-capping a chain moves the edge boundary, which can leave the tail
+        // as a child-only fork over a single embedded fork: merge that run into
+        // one edge, as a from-scratch build at the new boundary would.
+        if let ForkPayload::Child(Child::Embedded(table)) = &payload
+            && table.len() == 1
+            && let Some((first, record)) = table.iter().next()
+        {
+            return Box::pin(compact(store, prefix, first, record, stats)).await;
+        }
         return make_fork(prefix, payload, meta, child_count);
     }
     let head = prefix.get(..F::PLEN_MAX).ok_or(ApplyError::Internal)?;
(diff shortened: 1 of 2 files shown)


mfw78:
Superseded by #415, which fixes the same apply-vs-rebuild chain re-canonicalisation defect (this PR's #320 and the fuzz-found #414 are the same underlying bug). #415 flattens the child-only chain explicitly in `compact` rather than via `chain`/`compact` mutual recursion, absorbs this PR's re-capped-tail regression test verbatim, and additionally widens the long-key apply proptest (400 -> 800 bytes) so the three-fork-chain regime is reachable. Verified both ways: #415's fix passes this PR's #320 case, and this PR's fix passes #414's 511/766-byte cases. Closing in favour of #415.

PR #475, opened 2026-07-24
feat(postage-issuer): add generic parallel-sign helpers over any issuer

### LINKED ISSUE #38: [FEATURE] Parallel-sign helper accepting any StampIssuer
## What problem does this solve?
`nectar-postage-issuer::sharded::sign_stamps_parallel` only works with `ShardedIssuer`. Consumers that use `MemoryIssuer` (the common case for smaller workloads where the atomic-bucket sharding is overkill) cannot use it, so they roll their own. isheika does this at `client.rs:1527-1577`: a serial `&mut prepare_stamp` phase followed by a `rayon::par_iter` over the prepared digests with a shared `&signer`, then calling `BatchStamper::stamp_from_signature` (currently a private `const fn`) as a static helper to reconstruct the wire stamp.

Every consumer needing parallel signing reinvents this pattern.

## Describe the solution
Expose a generic parallel-sign helper at the crate top level that takes any `StampIssuer` and a signer, internally splits into the serial-prep / parallel-sign phases, and yields `Stamp`s. Make `stamp_from_signature` `pub` so callers can use it directly if they need a different concurrency strategy.

## Alternatives considered
- Migrate everyone to `ShardedIssuer`: not appropriate for small workloads; atomics are wasted overhead.
- Two phases manually per consumer: what isheika and any future consumer end up doing.
- Document the recipe in the crate docs: better than nothing but still leaves every consumer copy-pasting.

## Implementation notes
- The split is naturally: `prepare_stamps(chunks: &[Chunk], issuer: &mut impl StampIssuer) -> Vec<StampPreparation>` (serial) followed by `sign_stamps_parallel(preparations: &[StampPreparation], signer: &impl Signer) -> Vec<Stamp>` (parallel via rayon).
- Make `BatchStamper::stamp_from_signature` `pub`.
- The existing `ShardedIssuer` path can stay; this is a complementary helper for the non-sharded case.

## Are you willing to PR this?
Maybe later.

AI Assistance: Claude Code used for analysis and authoring this issue.

mfw78:
## What

Adds a generic serial-prepare / parallel-sign path over any `StampIssuer`, in a new `crates/postage-issuer/src/prepared.rs` module gated behind the existing `parallel` feature.

`prepare_stamps` allocates a digest per address serially (any issuer needs `&mut`, one at a time), preserving input order. Stamp timestamps come from the injected `nectar-clock` source: `prepare_stamps_with_clock` takes any `Clock`, and the plain `prepare_stamps` defaults to the system clock, matching the pattern main established for `sign_stamps_parallel` / `sign_stamps_parallel_with_clock` and `BatchStamper::with_clock`.

`sign_prepared_parallel` fans the signing step out across a rayon `par_iter`, taking the same `Fn(&B256) -> Result<Signature, E>` signer shape as the existing sharded `sign_stamps_parallel`; an allocation failure passes through unsigned rather than being signed.

`stamp_parallel` combines both steps for the common case of one issuer plus one signer, with `stamp_parallel_with_clock` as the deterministic-timestamp variant.

All helpers are exported at the crate top level from `lib.rs`, alongside the existing `StampResult` type which is reused for the output.

CI now covers the `parallel` feature for `postage-issuer` in both `lint.yml` (clippy) and `unit.yml` (nextest), matching the existing pattern used for `postage`'s `serde` feature.

## Why

Closes #38.

The existing `sign_stamps_parallel` only works over `ShardedIssuerFor`, which has a fully concurrent allocation path. Other issuers (e.g. `MemoryIssuer`, `RingIssuer`) need `&mut` for allocation and so cannot use that path; this gives them an equivalent serial-allocate / parallel-sign helper without touching any existing signatures. Purely additive: no dependency or `Cargo.lock` changes beyond the `nectar-clock` dependency already introduced on main.

The clock threading follows the injected-time rework that landed on main: the default clock preserves the previous behaviour byte for byte, and the `_with_clock` variants make wire-visible stamp timestamps deterministic under test.

## Testing

Scoped battery re-run locally on the rebased branch (based on `origin/main` at the clock-injection head), Rust 1.94 via `nix develop`, `RUSTC_WRAPPER=sccache`, `CARGO_INCREMENTAL=0`. All green.

- `cargo clippy --locked --all-targets -p nectar-postage-issuer` (default and `--features parallel`, after `cargo clean -p`): pass, no warnings in the crate.
- `cargo rustc --locked -p nectar-postage-issuer --lib -- -D unused_crate_dependencies`: pass.
- `cargo nextest run -p nectar-postage-issuer --locked` (default and `--features parallel`): 59 and 68 passed respectively.
- `cargo test --doc -p nectar-postage-issuer --locked`: pass.

Seven tests cover: `MemoryIssuer` bulk signing, signature recovery, in-order `BucketFull` and signer-error pass-through, the two-phase `prepare_stamps` / `sign_prepared_parallel` split, `RingIssuer` genericity, and deterministic timestamps through `stamp_parallel_with_clock` with a `ManualClock`.

## AI Assistance

Implementation by Claude (Fable), review by Claude (Opus).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -57,6 +57,10 @@ jobs:
             # above never see them.
             - name: cargo clippy (postage serde)
               run: cargo clippy --locked --all-targets -p nectar-postage --features serde
+            # The parallel signing helpers sit behind the parallel feature,
+            # likewise outside the default-features pass.
+            - name: cargo clippy (postage-issuer parallel)
+              run: cargo clippy --locked --all-targets -p nectar-postage-issuer --features parallel
             # `unused_crate_dependencies` is enforced per-library via `cargo
             # rustc` (not `[workspace.lints]`) so the flag applies only to each
             # crate's own lib target — never to benches/examples/tests (which
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -103,6 +103,13 @@ jobs:
                   cargo nextest run \
                     -p nectar-postage --features serde --locked \
                     --no-tests=warn --no-fail-fast
+            # The parallel signing helpers sit behind the parallel feature,
+            # outside the default-features pass above.
+            - name: Run postage-issuer parallel tests
+              run: |
+                  cargo nextest run \
+                    -p nectar-postage-issuer --features parallel --locked \
+                    --no-tests=warn --no-fail-fast
 
     wasm:
         # The postage-usage client facade is meant to run in a browser. The
--- a/crates/postage-issuer/src/lib.rs
+++ b/crates/postage-issuer/src/lib.rs
@@ -112,6 +112,8 @@ mod dilute_handler;
 mod error;
 mod factory;
 mod issuer;
+#[cfg(feature = "parallel")]
+mod prepared;
 mod ring;
 mod sharded;
 mod sharded_ring;
@@ -149,5 +151,10 @@ pub use factory::{
 };
 
 // Parallel signing (requires parallel feature)
+#[cfg(feature = "parallel")]
+pub use prepared::{
+    StampPreparation, prepare_stamps, prepare_stamps_with_clock, sign_prepared_parallel,
+    stamp_parallel, stamp_parallel_with_clock,
+};
 #[cfg(feature = "parallel")]
 pub use sharded::{StampResult, sign_stamps_parallel, sign_stamps_parallel_with_clock};
--- a/crates/postage-issuer/src/prepared.rs
+++ b/crates/postage-issuer/src/prepared.rs
@@ -1,0 +1,283 @@
+//! Serial-prepare / parallel-sign helpers for any [`StampIssuer`].
+//!
+//! Index allocation needs `&mut` and stays serial; only the signing fans out
+//! across threads. [`ShardedIssuerFor`](crate::ShardedIssuerFor) additionally
+//! has a fully concurrent allocation path in
+//! [`sign_stamps_parallel`](crate::sign_stamps_parallel).
+
+use alloy_primitives::B256;
+use alloy_signer::Signature;
+use rayon::prelude::*;
+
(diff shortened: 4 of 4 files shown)


mfw78:
Follow-on direction so review effort lands in the right place: a stacked PR will front a bounded-window streaming pipeline (`StampPipeline`) as the single public many-chunk stamping entry and seal this PR's exports (`StampPreparation`, `prepare_stamps`, `sign_prepared_parallel`, `stamp_parallel`) behind it as internal machinery, so these exports are intentionally short-lived and should not be treated as published surface. Because the crate is published and 0.4.0 predates prepared.rs, the demotion is free only while these stay unreleased, which makes the stacked PR a release blocker for 0.5.0 once this merges. The stacked PR also assumes this branch is rebased onto main's clock-injection seam (#476/#480) so prepared.rs takes an injected clock rather than the free-function timestamp; no other changes are requested here.

mfw78:
Update superseding the earlier ordered-pipeline note: the follow-on surface is reshaped to unordered completion as the only mode, results yield as they complete with each StampResult carrying its ChunkAddress under a multiset 1:1 contract (exactly one result per input address instance), and the reorder ring, front-slot wait and input-order guarantee are deleted (head-of-line blocking behind a straggler signature disqualified ordered yield; variable-latency remote signers are first-class, and no consumer in nectar or vertex depends on result order). Everything else in the earlier note stands: this PR's four exports plus sign_stamps_parallel_with_clock remain intentionally short-lived internal machinery, the rebase onto main's clock-injection seam (#476/#480) is still required before merge, and the follow-on work must merge before 0.5.0 or the exports leak into the published surface. One change of plan: sign_stamps_parallel is now removed outright in the 0.5.0 train rather than deprecated, with the sharded &self-allocation specialization landing in the same train as its replacement and both benches ported in the removal car. The work is now tracked as epic #484 with sub-issues rather than a single follow-on PR.

Issue #471, opened 2026-07-24
primitives: versioned modern envelope over the ecies compat baseline


mfw78:
Rescoped with the settled design after the #481 review. The compat construction in `crates/primitives/src/ecies.rs` stays byte-frozen exactly as #481 lands it; nothing below touches that module.

## Construction

RFC 9180 HPKE, single suite, no agility: DHKEM(secp256k1, HKDF-SHA256), HKDF-SHA256, ChaCha20-Poly1305, domain label `nectar/env/v1`. Nectar-to-nectar traffic only; compat remains the default toward unknown readers.

Standards caveat: this KEM is draft-wahby-cfrg-hpke-kem-secp256k1 (requested codepoint 0x0016), an expired individual CFRG draft, not part of RFC 9180 and not in the IANA HPKE registry; 0x0016 is a de facto ecosystem convention only. Pin the draft by name and codepoint in module docs and carry nectar-owned conformance vectors; there is no registry to lean on.

## Seam

Sealed trait with two uninhabited marker types, not an enum. The no-downgrade rule rides the recipient handle's type parameter, so the scheme must be a type, not a runtime value; a tiny `SchemeId` (Copy, two variants) exists only as a report in the open result, never as a dispatcher. Heterogeneous collections get an edge-only `AnyRecipient` enum that matches once and calls the monomorphized path; the enum never enters the seam.

```rust
pub trait Scheme: sealed::Sealed + Sized {
    type Context<'a>: From<&'a Topic>;   // compat: Salt<'a>; hpke: Info<'a> ("nectar/env/v1" || topic)
    type Extra;                          // compat: EncryptionKey; hpke: Exporter
    type SealError: core::error::Error;
    const TAG: usize;                    // 0 for compat, 16 for hpke
    const ID: SchemeId;

    #[cfg(feature = "encryption")]
    fn seal(recipient: &Recipient<Self>, ctx: Self::Context<'_>, plaintext: &[u8], ct_len: usize)
        -> Result<SealedRecord<Self>, Self::SealError>;

    fn open(secret: &SecretKey, ctx: Self::Context<'_>, record: &Record<'_>)
        -> Result<Option<Opened>, OpenError>;
}

pub enum Compat {}  // thin adapter over ecies.rs, wire-frozen
pub enum Hpke {}    // the suite above, one sealed type, no agility
```

## Wire framing

[...]

PR #481, opened 2026-07-24
feat(primitives): add secp256k1 ecies module

### LINKED ISSUE #85: feat(primitives): secp256k1 El-Gamal/ECIES module (gates PSS and ACT)
Part of the layer-2 chunk primitives epic.

## Goal

Add a secp256k1 El-Gamal / ECIES module to `nectar-primitives`. This is the one cryptographic building block the layer-2 work needs that does not already exist, and it gates both PSS (trojan chunks) and ACT.

## Why here

The symmetric half (Keccak-256 CTR cipher) already lives in `chunk/encryption`. What is missing is the asymmetric half: ephemeral keypair generation, ECDH shared-secret derivation, and the key-derivation glue that feeds the existing cipher. Both PSS and ACT need it, so it belongs in `nectar-primitives` (proposed module `ecies`) rather than being duplicated.

## Construction

Encryptor (`recipient_pubkey`, `topic`/salt, padding):

1. Generate an ephemeral secp256k1 keypair.
2. ECDH shared secret = `DH(ephemeral_priv, recipient_pub)`.
3. Block key = `keccak(shared || topic)`.
4. Encrypt the padded plaintext with the Keccak-CTR cipher.

Decryptor: given the recipient private key and the ephemeral public key, derive the same key and decrypt. The 8-byte hint `keccak(key || topic)[0:8]` lets a recipient cheaply test a topic before attempting a full decrypt.

## Scope

- [ ] `ecies` module in `nectar-primitives` with encryptor/decryptor over secp256k1 + the existing Keccak-CTR cipher.
- [ ] Ephemeral key handling and ECDH consistent with the reference implementation (this is wire-critical: PSS chunks are decryptable cross-client).
- [ ] Hint derivation helper (`keccak(key || topic)[0:8]`).
- [ ] Keep the module wasm-friendly (no native-only crypto backend assumptions).
- [ ] Conformance vectors: encrypt with a fixed ephemeral key and recipient key, pin the ciphertext and hint.

## Dependencies

None. Blocks the PSS and ACT issues.

Part of #82.

mfw78:
## What

Adds a new `ecies` module to `nectar-primitives` (`crates/primitives/src/ecies.rs`) implementing reference-compatible secp256k1 ECIES: `shared_key` performs ephemeral ECDH via `k256::ecdh` and derives `keccak256(x || salt)`, with the shared x coordinate serialised big-integer style (leading zero bytes stripped) to match the reference client. `Hint` derives `keccak256(key || salt)[..8]` for cheap salt matching. `encrypt`/`encrypt_with`/`decrypt` wrap the existing Keccak-CTR cipher (`chunk::encryption::transcrypt_in_place`) with reference padding semantics: the keystream runs over the plaintext only, and any padding is appended as raw random bytes outside the keystream.

A typed `EciesError` covers the padding-length guard (`checked_sub`, no unwrap/panic in `src`), and `PrimitivesError` gained an `Ecies` variant. `k256`'s `ecdh` feature is now enabled for `nectar-primitives`.

## Why

Closes #85.

## Testing

Independent verification gate for `feat/primitives-ecies` (single commit `371b19c`, based on `origin/main`). All CI-mirror commands run locally under `nix develop` (Rust 1.94, matching the pinned CI toolchain) with the sccache wrapper from the dev shell.

- `cargo fmt --all -- --check` clean.
- `cargo clippy --workspace --all-targets --locked` passes; the only warnings are pre-existing and outside this diff (`chunk/type_tag.rs:110` use_self; `crates/testing/src/lib.rs:40-41` doc_lazy_continuation).
- `cargo clippy -p nectar-primitives --all-targets --locked --features encryption` clean, covering the `#[cfg(feature = "encryption")]` encrypt path.
- `cargo nextest run --workspace --locked`: 959 passed, 1 skipped, including the 10 `ecies::tests`, which pin four reference-client-derived vectors (basic, short salt, empty salt/plaintext, and a leading-zero-x edge case) plus roundtrip/error-path coverage.
- Independent reimplementation of the whole construction from scratch (separate secp256k1 scalar-mult, keccak, and Keccak-CTR keystream in Python) reproduced every pinned vector byte-for-byte, confirming the vectors are genuine for the construction specified in the issue rather than circular against this implementation.

## AI Assistance

Implementation by Claude (Fable); independent review by Claude (Opus).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1426,6 +1426,7 @@ dependencies = [
  "ff",
  "generic-array",
  "group",
+ "hkdf",
  "pkcs8",
  "rand_core 0.6.4",
  "sec1",
@@ -1777,6 +1778,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fda06d18ac606267c40c04e41b9947729bf8b9efe74bd4e82b61a5f26a510b9f"
 dependencies = [
  "arrayvec",
+]
+
+[[package]]
+name = "hkdf"
+version = "0.12.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b5f8eb2ad728638ea2c7d47a21db23b7b58a72ed6a38256b8a1849f15fbbdf7"
+dependencies = [
+ "hmac",
 ]
 
 [[package]]
@@ -2325,6 +2335,7 @@ dependencies = [
  "getrandom 0.4.2",
  "hybrid-array",
  "js-sys",
+ "k256",
  "keccak-batch",
  "nectar-marker",
  "nectar-testing",
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -29,6 +29,7 @@ alloy-signer-local.workspace = true
 ## rust crypto
 digest.workspace = true
 hybrid-array.workspace = true
+k256 = { workspace = true, features = ["ecdh"] }
 keccak-batch.workspace = true
 
 # derive macros
--- a/crates/primitives/src/ecies.rs
+++ b/crates/primitives/src/ecies.rs
@@ -1,0 +1,376 @@
+//! secp256k1 ECIES: ephemeral ECDH feeding the Keccak-256 CTR cipher.
+//!
+//! The cipher key is `keccak256(x || salt)` where `x` is the ECDH shared
+//! point's x coordinate serialised with leading zero bytes dropped, matching
+//! the reference client's big-integer encoding. Wire-critical: ciphertexts
+//! must decrypt cross-client.
+
+use alloy_primitives::Keccak256;
+use thiserror::Error;
+
+pub use k256::{PublicKey, SecretKey};
+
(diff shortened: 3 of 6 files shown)


mfw78:
Review response, pushed as 518c4eab: the salt is now a zero-cost Salt newtype rather than a bare byte slice. Why it was unbounded: the reference construction genuinely accepts arbitrary salts (the access-control derivations use one-byte salts, topics are 32 bytes, empty occurs), so a hard 32-byte type would break the planned access-control consumer; the conformance vectors pin the short shapes for exactly that reason. The trap was real twice over though: encrypt took two adjacent untyped slices so salt and plaintext could swap silently, and the arbitrary length hid that the keccak(x || salt) input is not length-delimited at the x boundary (x is stripped of leading zeros). Salt fixes the swap at compile time and makes short shapes an explicit Salt::raw escape with the caveat documented; 32-byte topics convert frictionlessly via From. Conformance vectors are byte-identical (the wrapper is transparent), 258 primitives tests green. Proper length domain separation belongs to the modern envelope (#471), not the compat baseline.

Issue #473, opened 2026-07-24
epic: file read-ahead ergonomics and read-path hardening


mfw78:
Land the read-ahead trio and harden the read paths. Independent of the primitives epic except for the shared coordinated breaking release.

1. Read-ahead: #404 folds into #402's PR (doc fix plus static throughput hint), then #403 (adaptive controller from walk occupancy). #403 takes its clock from `nectar-clock` (#470) so the core stays `no_std`-clean.
2. Breaking hardening: #409 (zeroize the serialized encrypted-reference buffers) and #410 (narrow the read paths to `ContentOnlyChunkSet` with the address check pinned into decode). Both ride the coordinated breaking release with primitives #317/#318/#319.
3. #405: wasm data sink over a browser file handle, independent of the rest.

Constraint carried as DoD: BMT-segment-granularity witness access must survive the epic, or reviving zkvm #306 later needs a codec change.

Done when the trio is landed, the read paths are narrowed, the encrypted-reference buffers are zeroized on drop, and the file share of the #230 suppression register is cleared and reported there.

mfw78:
Closing as drained. Every child of this epic is closed, and its thesis is superseded by the standards now recorded in `AGENTS.md`: one concurrency substrate (`FuturesUnordered` plus the governor), layer discipline, and the packaging plan. The machinery this epic was written against no longer exists in the tree.

Remaining work in this area is tracked by the production plan epics: #668 wire-format vectors, #677 crate-set decruft, #684 type safety and the error contract, #697 CI gates, #705 tracker and specifications, #714 release readiness, #693 deferred capability gaps.

PR #537, opened 2026-07-26
feat(postage-issuer): add StampedPut store decorator

### LINKED ISSUE #487: postage-issuer: StampedPut store decorator over the split seam
## What

`StampedPut<I, Sg, P>` in nectar-postage-issuer: implements `ChunkPut` facing up and requires `P: PutStamped` facing down, so every existing persistence call site in file, manifest and mantaray gets stamping with zero code change in those crates.

## API and bounds

- One generic `impl ChunkPut<AnyChunkSet<B>>` covers file (any B), manifest (B = 4096) and mantaray (`StandardChunkSet = AnyChunkSet<DEFAULT_BODY_SIZE>`). `Clone + 'static` for the split engine's clone-per-put dispatch, with clone-SHARED state (`Arc<Mutex<..>>` native, `Rc<RefCell<..>>` under unsync): one issuer watermark and one issued map across all clones; a value-cloned issuer is a duplicate-index bug and the docs say so.
- The put future is MaybeSend and tolerates the split's synchronous noop-waker first poll: it allocates via `prepare_stamp` on first poll, signs through the shared `sign_task` engine, then awaits `inner.put_stamped(StampedChunk::new(chunk, stamp))`.
- Delegation impls so combined-bound sites work unchanged: `ChunkGet` (Trust = inner's trust) and `ChunkHas` pass through. Per-site inner bounds, stated in docs: put-only sites (Split, collect_into, TokioWriter, split_read_at, builder, save) need `P: PutStamped`; editor commit and manifest apply need `P: PutStamped + TrustedGet + ChunkHas`, and a pure network sender cannot honestly claim `TrustedGet`, so those sites take a local or tee'd inner (see the postage sub-issue).
- Error: a composite `ChunkPut::Error` over StampError, SigningError and P::Error, surfacing through `SplitError::Put`; plus an issuer remaining-capacity accessor so callers can preflight a nearly-full batch instead of aborting mid-tree (per-item BucketFull is terminal at every wrapped site).

## Per-address idempotence (normative, not optional)

Without it the decorator deterministically fails repetitive data that uploads fine unstamped: a 1 MiB zero-filled region splits into 256 identical leaves with one address, hence one bucket; bucket capacity at depth 20 / bucket depth 16 is 16, so the 17th duplicate put hits BucketFull and the split's poison fuse aborts the whole upload. The clone-shared issued map (address to stamp) is consulted before allocation; a duplicate put reuses the already-signed stamp, or short-circuits once the inner put has succeeded. Memory bound ~145 B per unique address, with a bound/off switch for huge uploads; `ChunkHas` preflight remains the optional cross-session dedup layer on top (a store-level preflight alone cannot close this, since two puts of one address can be in flight in one window). Pinned: split a zero-filled buffer larger than bucket-capacity leaves through the decorator and assert one stamp, one inner put, success rather than BucketFull.

## Composition rules (normative docs)

- Transport retries compose below `PutStamped` (inside put_stamped, reusing the already-signed stamp), never by re-invoking the decorator's `ChunkPut`; the issued map makes an accidental re-put idempotent anyway, and primitives' retrying decorator already delegates put exactly once.
- The decorator stamps per put, not per reachable chunk: mantaray save and editor commit skip already-referenced nodes, so a wrapped re-commit stamps only newly-put chunks; whole-tree restamping is the pipeline's direct entry.
- Wrapping a purely local store burns indices for chunks that may never reach the network, and the split's poison fuse makes the only retry a full re-split; compose a presence filter upstream where that matters.

## Split integration (the #473 edge)

- `Split<StampedPut<..>>` needs zero crates/file changes: dispatch already folds completions in any order and the PutWindow InFlight set tolerates unordered completion natively.
- PutWindow slots double as sign-plus-put slots (default 16, 16x below SignWindow's validated 256): docs must say widen the put window towards `SignWindow::default()` when wrapping with a slow signer; a wrapped-split bench is this issue's acceptance gate, and a dedicated sign stage between the hash fan and the put window is the contingency if it reproduces the w=64-class degradation.
- collect_into is a serial seam: drain awaits each outer put to completion and the relay-backed split never engages its window (relay puts complete on the synchronous first poll), so a 50 ms remote signer costs one round-trip per chunk, roughly 50 s for a 4 MiB file versus roughly 3 s through an owned-store split at window 16; docs steer decorator users to `Split::new`/collect with an owned clone.
- `debug_assert!(rayon::current_thread_index().is_none())` on the put future's first poll (it runs on the driver thread, so this exactly catches pool-thread-driven splits); spawn_blocking guidance mirrored from the pipeline.

## Dependencies

After the pipeline sub-issue (shared `sign_task` engine) and the postage PutStamped sub-issue. Rides 0.5.0; the wrapped-split bench lands under the #473 file epic umbrella.

mfw78:
## What

Adds `StampedPut<I, Sg, P, C = SystemClock>`, a store decorator that stamps every chunk at put time, implementing `ChunkPut<AnyChunkSet<B>>` facing up over a `P: PutStamped<B>` sink facing down.

The type lives in `crates/postage-issuer/src/pipeline/stamped_put.rs`, inside the `pipeline` module so it consumes the existing `pub(crate)` sign engine with no visibility changes. `pipeline/mod.rs` and `lib.rs` are touched only for exports.

`Clone` shares state: an `Arc<Mutex<State>>` under native builds, an `Rc<RefCell<State>>` under `wasm32`/`target_os = "none"`/`unsync`, with the module absent entirely on hosted no-std builds without those cfgs. Shared state holds one issuer watermark and one issued map (`BTreeMap<ChunkAddress, Issued>`), sized by an `IssuedBound` (`Unbounded` / `AtMost` / `Off`).

The put path is an async fn built to tolerate the split's synchronous noop-waker first poll: it checks a `debug_assert!` that it is never driven on a rayon worker, then consults the issued map before allocating a stamp index, under one lock/borrow that is never held across an await.

A follow-up commit on the branch adds an `Issued::Delivering` state plus a `DeliveryGuard` so concurrent in-flight duplicate puts of the same address share a single sink delivery instead of racing.

`Cargo.toml` gains a `nectar-testing` dev-dependency and switches `nectar-file` to its `std` feature for the decorator's regression tests.

## Why

Closes #487. Stamping today happens as a separate pass over already-collected chunks; this decorator lets any existing `ChunkPut` call site stamp at put with zero changes to that call site, by wrapping the sink instead of threading stamping through it.

## Testing

Independent gate on `feat/stamped-put-decorator` @ `bde6e424`, based on `origin/main`. Changed files: `Cargo.lock`, `crates/postage-issuer/{Cargo.toml, src/lib.rs, src/pipeline/mod.rs, src/pipeline/stamped_put.rs}`. Zero changes under `crates/file`.

All commands run under rust 1.94 (via `nix develop`, matches CI pin) with a shared `CARGO_TARGET_DIR` and `RUSTC_WRAPPER=sccache`:

- Default tests: 87 passed, plus doctests.
- Parallel tests (`parallel` feature): 88 passed, including the concurrency test.
- Unsync build: green.
- `--locked` build: green.
- Strict clippy `--all-targets`: clean (only `nectar-testing` dependency warnings).

Verified the single-lock concurrency protocol has no lost-wakeup or deadlock, that cfg gating is symmetric with `nectar-marker`'s `MaybeSend`/`MaybeSync` and guarded by a `parallel` + `unsync` `compile_error!`, that issued-map keying matches `chunk.address()` with no wire or behaviour drift, and that the zero-region `== 3` assertion in the bucket-capacity test is correct.

Also re-checked the prior gate's blocker: the intermittent `pipeline::stamped_put::tests::zero_region_larger_than_bucket_capacity_stamps_once` failure (`seen.len()` 4 vs 3 under the `parallel` feature) is fixed by the `Issued::Delivering` / `DeliveryGuard` commit.

Closes #487.

## AI Assistance

Implemented by fable, reviewed by opus, per the org AI-assistance contract (nxm-rs/.github CONTRIBUTING.md).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2699,6 +2699,7 @@ dependencies = [
  "nectar-marker",
  "nectar-postage",
  "nectar-primitives 0.4.0",
+ "nectar-testing",
  "proptest",
  "rand 0.10.1",
  "rayon",
--- a/crates/postage-issuer/Cargo.toml
+++ b/crates/postage-issuer/Cargo.toml
@@ -33,8 +33,10 @@ alloy-signer-local = { workspace = true, optional = true }
 rayon = { workspace = true, optional = true }
 
 [dev-dependencies]
-# Referenced by the `arbitrary` feature.
-nectar-file = { workspace = true }
+# Referenced by the `arbitrary` feature; the decorator regression splits
+# through the streaming engine.
+nectar-file = { workspace = true, features = ["std"] }
+nectar-testing = { workspace = true }
 # The upload pipeline bench splits through the streaming engine and the
 # rayon batch ingest.
 proptest = { workspace = true }
--- a/crates/postage-issuer/build.rs
+++ b/crates/postage-issuer/build.rs
@@ -1,0 +1,13 @@
+//! Emits the `multi_thread` cfg alias.
+
+fn main() {
+    // Send/Sync bounds apply; wasm32, bare metal and the `unsync` feature
+    // relax them, mirroring the marker traits.
+    println!("cargo::rustc-check-cfg=cfg(multi_thread)");
+    let wasm32 = std::env::var("CARGO_CFG_TARGET_ARCH").as_deref() == Ok("wasm32");
+    let bare = std::env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("none");
+    let unsync = std::env::var_os("CARGO_FEATURE_UNSYNC").is_some();
+    if !wasm32 && !bare && !unsync {
+        println!("cargo::rustc-cfg=multi_thread");
+    }
+}
--- a/crates/postage-issuer/src/lib.rs
+++ b/crates/postage-issuer/src/lib.rs
@@ -12,7 +12,8 @@
 //!   and yields results unordered, tagged by address.
 //! - [`BatchStamper`]: the one-off entry for stamping a single chunk;
 //!   [`BatchStamper::into_parts`] moves an issuer between the doors.
-//! - Stamping at put lands later as a store decorator over the same engine.
+//! - [`StampedPut`]: the store decorator; wraps a stamped sink so every
+//!   `ChunkPut` call site stamps at put.
 //!
 //! # Immutable and mutable issuance
 //!
@@ -169,6 +170,13 @@ pub use stamper::{BatchStamper, Stamper};
 // The streaming stamp pipeline; its sign window is the kernel window.
 pub use nectar_kernel::Window;
 pub use pipeline::{Eip191, SignPrehash, StampPipeline, StampResult, Stamped};
+#[cfg(any(
+    feature = "std",
+    target_arch = "wasm32",
+    target_os = "none",
+    feature = "unsync"
+))]
+pub use pipeline::{IssuedBound, StampedPut, StampedPutError};
 
 // Mutable (ring) issuing with a type-state reservation guard
 pub use ring::{Reservation, Reserved, RingIssuer, RingIssuerFor, Unreserved};
--- a/crates/postage-issuer/src/lib.rs
+++ b/crates/postage-issuer/src/lib.rs
@@ -170,12 +170,7 @@ pub use stamper::{BatchStamper, Stamper};
 // The streaming stamp pipeline; its sign window is the kernel window.
 pub use nectar_kernel::Window;
 pub use pipeline::{Eip191, SignPrehash, StampPipeline, StampResult, Stamped};
-#[cfg(any(
-    feature = "std",
-    target_arch = "wasm32",
-    target_os = "none",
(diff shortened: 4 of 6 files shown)


mfw78:
Post-merge note for the record: #487 named a wrapped-split throughput bench as this work's acceptance gate (does stamping through the 16-slot PutWindow starve the upload under a slow signer, the w=64-class degradation). This PR shipped the decorator plus its deterministic idempotence test (the zero-buffer BucketFull case); the throughput bench is tracked separately as #540 (under the #473 file epic umbrella per #487), which also sizes the widen-the-put-window docs guidance and gates the dedicated-sign-stage contingency. The decorator is purely additive (Split<StampedPut> needs zero crates/file changes), so no existing upload/split benchmark can regress from it; #540 is new characterization, not a regression check.

PR #566, opened 2026-07-27
manifest: window builder and apply puts

### LINKED ISSUE #555: manifest: window the 1.0 builder node/segment puts and drop the keymap clone
#553. The 1.0 KV build (`crates/manifest/src/builder.rs:187-249,454-601`) assembles streaming (Frame stack, O(depth) open nodes) but awaits EVERY node emit and EVERY segment put inline - one put in flight per publish. Fold `emit_node`/`resolve`/`spill_node` put awaits onto a bounded put window in the `CommitPolicy` shape (`mantaray editor.rs:356-616` is the in-repo template); sibling spills + segment puts are order-free (content-derived addresses). Extract the Split engine's admit/dispatch/step trio as the shared bounded-put-sink rather than a second hand-rolled copy. Also kill the gratuitous whole-key-map -> `Vec<Item>` clone in `build()` (191-199): consume the map / accept a pre-sorted iterator, making the doc header's memory-bounded claim true of the input too.

Pattern: fold-onto-Driver + make-streaming. Guard: the 1.0 byte-vector/CDC anchors + O(depth) membound green. Benefits from #554.

mfw78:
## What

Adds a bounded put window to `nectar-manifest`'s build and apply paths.

`builder.rs` gains a `pub(crate)` `PutSink<'a, S>`, built from kernel `Admission` + `InFlight<'a>` + `Window`, that dispatches sealed chunk puts into an overlapping window instead of awaiting each inline, with a final drain before the root returns.
`emit_node`, `resolve`, `spill_node`, `put_payload`, and `put_counted` seal chunks (content-derived addresses) and dispatch through the sink.
`Builder::build(self, store)` now consumes the sorted key map so its `BTreeMap` entries move into the descent array without a clone, and `Builder` gains `with_put_window(Window)` (defaulting to `Window::DEFAULT`), mirroring the mantaray editor's commit window.
`nectar_kernel::Window` is re-exported from the manifest crate root.

`apply.rs`, the shared trio's other caller, is carried onto the same seam: `PutSink` is threaded through `apply_forks`/`reconcile`/`split`/`descend`, reads go via `puts.store()`, and the apply drains before its new root returns.

Closes #555

## Why

Node and segment puts are order-free once addresses are content-derived at seal time, so awaiting each one inline serialises work that has no ordering dependency. A bounded window lets independent spills overlap while keeping the retained put count capped, without introducing dyn dispatch or spawned tasks.

## Testing

- `cargo clippy --workspace --all-targets --locked`: clean (only pre-existing `doc_lazy_continuation` warnings in untouched `crates/testing`).
- New integration test `spill_puts_overlap_up_to_the_put_window` (`crates/integration-tests/tests/manifest/builder.rs`) drives the build against a gated store to observe that puts genuinely overlap up to the configured window, that the window caps peak concurrency, that every put settles before the root returns, and that the windowed root equals a serial build's.
- Existing `apply_equals_rebuild` oracle re-run and passing.

Independent adversarial review found no correctness, wire-byte, panic, lint-suppression, or house-rule defects: addresses are content-derived locally before dispatch so bytes are order-independent, drain and full-window polls propagate the first put failure with no silent drops, and apply reads only pre-existing old-tree nodes so there is no read-after-write hazard against in-flight puts.

## AI Assistance

Implemented by fable, reviewed by opus.

### DIFF
--- a/crates/integration-tests/tests/manifest/builder.rs
+++ b/crates/integration-tests/tests/manifest/builder.rs
@@ -4,14 +4,20 @@
 //! buffers track the trie depth rather than the key count, and the files path
 //! splits through BMT and references the stored roots.
 
+use core::convert::Infallible;
+use core::task::Poll;
+
 use anyhow::{Context, Result, anyhow, ensure};
 use bytes::Bytes;
 use nectar_manifest::{
     BuildStats, Builder, Child, Entry, ForkPayload, ForkTable, Key, KeyId, Metadata, Node, NodeGet,
-    Prefix, RootExtension, V1, build_files,
+    Prefix, RootExtension, V1, Window, build_files,
 };
-use nectar_primitives::{ChunkAddress, ChunkOps, ChunkRef, ContentGet, MemoryStore};
-use nectar_testing::{run, split_whole};
+use nectar_primitives::store::ChunkPut;
+use nectar_primitives::{
+    Chunk, ChunkAddress, ChunkOps, ChunkRef, ContentGet, MemoryStore, Verified,
+};
+use nectar_testing::{Drive, GateStore, run, split_whole};
 
 const fn ref32(byte: u8) -> ChunkRef {
     ChunkRef::new(ChunkAddress::new([byte; 32]))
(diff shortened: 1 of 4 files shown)


mfw78:
Superseded: this was built on nectar_kernel::InFlight, which #571 (PR #577) deleted in favour of FuturesUnordered. Redoing #555 fresh off current main on the FuturesUnordered substrate (mirroring the migrated relay/store put-window pattern).

Issue #618, opened 2026-07-28
phase 4: dissolve the benches and examples crates; codspeed to criterion


mfw78:
Part of the production plan, `docs/PRODUCTION-PLAN.md`, stage 2. Independent; based on main.

Re-scoped 2026-08-12: this issue is now the crate dissolution and the benchmark harness swap only. The `crates/integration-tests` slimming moved to its own issue, because the two are separately reviewable and the crate-set change should settle the workspace member count on its own.

## What

Move each `[[bench]]` and `[[example]]` out of `crates/benches` and `crates/examples` into its logical host crate as a dev-dependency target, then dissolve the two aggregator crates. A `[[bench]]` or `[[example]]` declared in its host is invisible to a normal build, so this costs nothing at build time and removes two workspace members.

Keep `dump_deployments` runnable. It lives at `crates/examples/examples/dump_deployments.rs` and `.github/workflows/upstream-addresses.yml` invokes it; repoint that workflow at `-p nectar-contracts`.

Drop the dormant codspeed shim in the same change. `Cargo.toml:144` declares `criterion = { package = "codspeed-criterion-compat", version = "2.10.1" }`, and no workflow runs `cargo bench`, so the compat layer buys nothing. Replace it with plain `criterion`. This is coupled to the dissolution rather than separable: codspeed is a dependency of `crates/benches`, which this issue deletes.

## Kills

Two workspace members, taking the tree from 22 to 20. The codspeed dependency and its transitive `uuid` and `tinytemplate` edges, which currently sit in the licence and advisory graph for a tool nothing runs.

## Gate

`cargo metadata` reports 20 members and the number is written down in the pull request body. `cargo build --workspace --all-targets --locked` succeeds, proving every moved target still builds in its new host. `cargo bench --no-run` builds the moved benches. `upstream-addresses.yml` runs `dump_deployments` from `nectar-contracts`. No crate named in `.github/workflows/nostd.yml` or in the `lint.yml` crate list is left naming a dissolved member.

mfw78:
[...]

PR #648, opened 2026-07-29
refactor(ldb,mantaray,manifest)!: model the manifest KV surface on HashMap idioms

mfw78:
## What

Models the manifest key-value surface on `HashMap` idioms, on a wire that stays byte-identical to the reference client's mantaray v0.2.

The read handle is `MapView` and the write handle is `MapWriter`, bound by `Manifest::at` and `Manifest::edit`. The verbs are `get`, `insert`, `remove`, `contains_key`, `iter`, `range`, `floor`, plus the two manifest additions `dir` and `load`. `insert` replaces the whole binding, and `remove` is exact-key on both formats, exactly as `HashMap::remove` is: the key's own value and metadata go, a key with children keeps every one of them, a childless leaf is pruned, and removing an unbound or absent key is a no-op. mantaray's legacy boundary remove lives on as `remove_subtree`, for the pinned 0.3.0 differential alone.

Content keys are stored bare and verbatim. `ManifestPath::from("index.html")` is the bytes `index.html` on both formats, which is what the reference client writes, so mantaray's image is byte-identical to bee v0.2 and `nectar-ldb` keeps its native encoding.

The manifest's own configuration is an explicit option-typed API and never a key:

- `MapView::index_document()` and `MapView::error_document()` answer `Option<ManifestPath>`, `None` when unset, over one `site_config()` read that serves both.
- `MapWriter::with_index_document(path)` and `MapWriter::with_error_document(path)` are chainable and take `impl Into<Option<ManifestPath>>`, so the same verb sets and clears.
- Each lands in the format's native root slot. mantaray merges the metadata onto the `"/"` node, which binds the zero address and therefore carries no entry, which is the layout `pkg/api/bzz.go` writes and its three `Lookup(RootPath)` reads expect. `nectar-ldb` merges its root manifest metadata.
- The content map surfaces neither that slot nor an empty key. The empty path is a listing prefix, so `dir` lists the top level with it while every other verb answers absent.

## Why

Two rounds of this branch rooted the canonical key at the separator, so `"/"` named the manifest and `"/index.html"` a file below it. That was wrong twice over, and this PR retracts both claims.

It broke the wire. Bee passes the path straight to `trie.Add` (`pkg/manifest/mantaray.go`), so a content key is bare. A rooted key is a different byte image for the same content, which no reference client reads.

It neutered the gate that would have caught it. The differential against the pinned 0.3.0 crate had been rescoped to feed both sides the same rooted keys, which measured the merge algorithm and no longer measured the wire.

Modelling the site documents as metadata on a root key was the same mistake in the API. Bee stores them as metadata at `RootPath = "/"` beside a zero-address entry (`pkg/manifest/manifest.go`, `pkg/api/bzz.go`), and the empty path is the structural trie root, which bee never keys site config at. Neither is a content key, so neither belongs in the map.

## Testing

- The randomized differential against the pinned 0.3.0 release is deleted, together with its `mantaray-old` dev-dependency. That release is nectar's own, so the differential compared this crate against its own earlier behaviour rather than against the network, and the 0.3.0 oracle saves one chunk per node so it cannot reach a multi-chunk root at all. The wire pin is now `tests/mantaray/bee_vectors.rs`, which asserts three roots the reference client produced, including that multi-chunk case. The properties the differential also carried survive in the crate: `split_commit_matches_the_fresh_replay` and `clean_ancestor_hazard_regression` in `crates/mantaray/src/editor.rs`, and randomized coverage through the `mantaray_editor_differential` and `mantaray_view_differential` fuzz targets against the shared oracles, replayed on stable from the committed seeds.
- New bee-layout gate (`tests/mantaray/bee_layout.rs`) asserts the stored image below the seam. Content keys are bare, nothing sits under a rooted spelling, the site documents are metadata on an entry-less `"/"` node, the structural root binds nothing, and the seam's setters and the trie's own `set_index_document` produce one root.
- The cross-format gate is rescoped from the root key to the configuration (`tests/manifest/root_config.rs`). It pins that both formats agree on `index_document`, `error_document`, every set and clear shape, that clearing the last document restores the content-only root, that no map verb reaches the slot, and that exact-key remove agrees across formats.
- The erased seam round-trips `dyn_site_config` and `dyn_set_site_config`, and each format's native slot is asserted below the seam.

A compression pass derives the hand-rolled `ReservedKey` and `ManifestPath` impls through thiserror and derive_more, and dedupes the shared test fixtures, with no assertion or behavior pin changed; the delta versus main is now 49 files changed, 4307 insertions(+), 1105 deletions(-).

Gates, all green: `cargo clippy --workspace --all-targets --locked -- -D warnings`, plus the feature passes for `nectar-ldb`, `nectar-mantaray`, `nectar-file` and `nectar-integration-tests`; `cargo nextest run --workspace --locked` at 1209 passed; `cargo test --doc --workspace --locked`; no_std and wasm32 checks for the touched crates on `riscv64imac-unknown-none-elf` and `wasm32-unknown-unknown`; `cargo check --locked` in the fuzz workspace; `cargo fmt` clean.

A follow-up commit on this branch (`docs(agents,tests): name the reference vectors as the wire pin`) corrects the four sites that still named the deleted differential as live: two `AGENTS.md` lines, the module docs of `bee_layout.rs` and `manifest/root_config.rs`, and the orphaned `legacy_differential.proptest-regressions`. `AGENTS.md` now records that `remove_subtree` is consumed by the editor oracle in `crates/mantaray/src/oracles.rs`, so it is not deleted as unused.

## AI Assistance

Implement: claude-opus-5
Compress: claude-fable-5


mfw78:
Agent pre-review, 2026-08-11. Verdict: **land, with one follow-up**.

Wire verified intact. Reference vectors 1 and 2 reproduce byte-identically on the pinned `nectar-mantaray 0.3.0` oracle, there is no ldb codec hunk and no `DERIVE_TAG` hunk, and content keys carry no leading separator, so the two retracted rounds recorded in the body really are retracted.

## Follow-up, not a blocker

`crates/integration-tests/tests/mantaray/legacy_differential.rs` (425 lines) and its pinned `0.3.0` dev-dependency are deleted, but four surviving references still assert that gate exists: `AGENTS.md:25` and `AGENTS.md:27`, and the module docs of the two new test files. The Testing section of this PR body is also false at head: it says the legacy differential is restored and pins the v0.2 wire byte for byte against the 0.3.0 oracle. `legacy_differential.proptest-regressions` is orphaned in the tree and even gains a new `cc` line.

The swap is largely defensible: epic #651 already blesses `bee_vectors.rs` as the pin, and vector 3 exercises a multi-chunk root the 0.3.0 oracle cannot save at all. But three static point vectors do not replace a randomized proptest differential, the submission-order permutation cases, the split-commit-boundary root equivalence, or the failing-op-index parity.

The `AGENTS.md` half is the sharper risk, because it is standing instruction for future contributors and it misnames `remove_subtree`'s only real consumer, which is now the in-crate fuzz corpus at `crates/mantaray/src/oracles.rs`. An agent trusting that file could delete `remove_subtree` and invalidate the committed corpus.

Cleanup is mechanical. The one real decision: restore the randomized differential as a nightly job against the pinned oracle, or record in #651 that the three static vectors are the accepted pin.

mfw78:
Follow-up tracked: #722.

This pull request deletes the randomized differential against the pinned 0.3.0 release and its `mantaray-old` dev-dependency, and replaces the wire pin with three reference-client roots in `tests/mantaray/bee_vectors.rs`. That trade is deliberate: the 0.3.0 release is nectar's own, so the suite compared this crate against its own earlier behaviour, and the oracle saved one chunk per node so it could not construct the multi-chunk root the new vectors cover.

Three properties do not survive the swap: failing-operation-index parity, submission-order permutation against an independent implementation, and randomized cross-checking against a construction that is not this one. #722 restores them from a written specification rather than from a pinned release, so the oracle cannot age out of the property space the way a release oracle does.

Scope note: #618 previously listed the 0.3.0 oracle as something to retain when slimming `crates/integration-tests`. That line is corrected.

PR #649, opened 2026-07-29
feat(manifest-sim): two-arm mantaray 0.2 vs ldb benchmark harness

mfw78:
## What

This adds `crates/manifest-sim` (package `nectar-manifest-sim`, `publish = false`), a two-arm measurement harness that runs mantaray 0.2 (`nectar-mantaray` through the production `nectar-loadsave` seam) against mantaray 1.0 (`nectar-ldb`) over one corpus, one key order and one store implementation, and writes one JSON result document plus a markdown rendering of it.

The crate is the renamed and extended `crates/ldb-sim`. The binary `manifest-perf` has two modes. `manifest-perf run` measures every `(corpus, scale)` and writes the document; `manifest-perf render --in <json> --out <md>` reads it back and prints the tables. Flags are `--out`, `--scales`, `--max-mantaray-scale` and `--build-samples`, with `--in` on the render side.

The seam is one `Arm` trait that both formats implement over their own instrumented `CountingStore`, so the comparison is between counters and never between shared state. Each measured operation returns its capability class beside its cost, so an emulated 0.2 path is labelled at the point of measurement and the label rides through to the rendered table.

## Why

This reproduces the whitepaper methodology of TR-004-A in tree, so the numbers behind the mantaray 1.0 comparison can be regenerated from a commit instead of quoted from a document.

Store fetches, puts and bytes are the currency and wall-clock is illustrative. A capability gap is a null with a recorded reason, never an estimate. The 0.2 arm runs up to `--max-mantaray-scale` (default 1e5) because its editor commit materialises the whole trie in RAM; above the cap every 0.2 cell is a null with that policy as its reason and no multiplier exists there.

The reproduced axes are the capability matrix, 2.1 storage and slot utilisation and embedding, 2.2 get hops with the modelled RTT columns including the 100 ms mobile column, 2.3 prefix listing fair and pessimal, section 3 floor, ceiling and range multipliers, 6.1 the build frontier, 6.2 write amplification against K, and 4.3 pagination.

Three findings correct the whitepaper for the crates actually in tree, and the harness reports the crate in front of it rather than the historical one. The current 0.2 cursor streams in documented path order rather than unordered. Its `after` bound prunes, so ceiling is seek-grade rather than O(N). Its multi-op commit does amortise, so batched write amplification is measured rather than assumed flat. Floor stays the honest asymmetry: no public-API path seeks backwards, so it walks to rank.

## The new build wall-time axis

This adds one first-class axis the whitepaper did not have: a timed cold-pass build per (arm, corpus, scale), one untimed warmup then `--build-samples` timed passes over a fresh plain `MemoryStore`, reporting mean, min and keys per second.

It is deliberately fenced. It lives in its own `wall_time` document section, never in the deterministic one; it runs over a plain store so no counter atomic sits in the timed path; and the cold-pass caveat rides every cell verbatim and prints above the rendered table. Wall-clock never becomes the headline currency and no rendered speedup mixes fetches with nanoseconds.

Criterion benches were weighed and rejected for this axis. Apples-for-apples needs one corpus and one store, which two per-crate criterion targets cannot share without a dev-dependency cycle; a 1e6-key build fights criterion's iteration model; and one JSON document with one provenance block is the discipline the comparison rests on.

## Determinism and rendering

The document splits into a `deterministic` section that is bit-identical across runs under the master seed and a `wall_time` section that varies by design. `SOURCE_DATE_EPOCH` pins the timestamp and the git branch and commit are stamped into `meta`.

The renderer is Rust, not Python. It never measures, never fills a gap, never averages across corpora and never extrapolates a capped arm from a smaller scale. A missing figure prints as `--` and its recorded reason becomes a footnote. Two layout rules carry the findings: every table is keyed by corpus, and every multiplier prints beside the 1.0 absolute cost it divides, because a ratio that widens with N and a 1.0 floor that grows with depth are separate findings. Write amplification puts K on the row axis so the K=1 row can never be read apart from the curve that explains it.

## Testing

`cargo fmt --all --check` is clean. `cargo clippy -p nectar-manifest-sim --all-features --all-targets --no-deps -- -D warnings` is clean. `cargo clippy --workspace --all-targets --locked` shows only the pre-existing `nectar-testing` doc warning, which is out of scope here. `cargo test --doc -p nectar-manifest-sim` passes.

`cargo nextest run -p nectar-manifest-sim` is 30 green in 26 seconds with no slow test. The suite includes the byte-identity gates that serialise the deterministic section twice at 1e3 and assert that nothing the wall-time lane produces is inside it, the acceptance gates each metric module owns, and the renderer gates that a null prints as `--` with a footnoted reason and that the K=1 row stays inside the sweep table.

A round trip at `--scales 1000 --build-samples 1` writes the JSON, deserialises it back through the render mode, and prints every capped 0.2 cell as `--` with its policy reason footnoted.

## Rebased onto #644

The branch predated #644 (the `nectar-manifest` trait crate, now on main) and carried a rustfmt-only commit that conflicted with #644's edits to `crates/mantaray/src/editor.rs`. The manifest-sim work is now cherry-picked onto current main without that style commit, so the diff is the benchmark crate and nothing else. The remaining workspace rustfmt drift predates this branch and is owned by #644, not corrected here.

## Rendered-output honesty fixes

A lens sweep found the rendered markdown was less honest than the JSON behind it. Every gap below now has a test.

The document splits by audience. A `User experience` group holds the capability matrix, storage, hops, listing, ordered operations, cursor rounds, pagination and subtree serve. A `Developer experience` group holds V1Read against V1, the build frontier, write amplification and build wall-time. No table sits in both, so neither audience reads the other's number as its own.

Every table whose cells come from a measured operation now carries a capability legend beneath it. An unmarked figure is a native primitive of the format; a marked one names the emulation and its cost class. The labels were always in the JSON and the markdown dropped them.

The 1.0 ceiling is classed honestly. 1.0 has no dedicated ceiling primitive, so the arm composes `range(key, MAX)` with one `next()`, and that cursor fills a read-ahead window of speculative child fetches on its first poll. The measured absolute is therefore an upper bound on a dedicated seek, not the cost of one. The label rides every ceiling measurement and the capability matrix, and the section prints the bound beside the number. No new seek primitive is implemented: the fix is disclosure, and the measured figure is kept.

The 1.0 pessimal cells read as `native (no 1.0 fallback)`. They repeat the native cost because 1.0 has no degraded path, and the prose no longer implies a whole-manifest walk was measured on the 1.0 side.

The whitepaper divergences are a was-and-is table rather than a prose blob, so the corrections to row 7 (ceiling, was unsupported O(N), is an after-bound pruned seek) and row 10 (ordered iteration, was unordered with O(N) memory, is a streaming walk in documented path order) are rendered findings and not silent absorptions.

Three measured lanes now reach the markdown: the parallel cursor rounds (whitepaper 4.1), the V1Read profile (4.2) and the subtree handoff. They were populated in the JSON and never rendered.

Minors: the section-3 multiplier rows are ordered so each op's N-series ascends; the caveat and legend print once per corpus table rather than once per document; the `OrderedOpCell` fair and pessimal rustdoc said mean and held an aggregate; the resume-walk emulation label reaches the 4.3 column; and a footnote now matches the exact field it prints instead of borrowing an unrelated gap the same arm recorded.

## The uniform 1e3 seek

Review flagged that on uniform 1e3 the 1.0 floor and ceiling each charge 39 fetches, which is exactly that arm's resident chunk count, and asked whether a whole-tree walk was masquerading as an O(depth) seek.

It is legitimate at that shape. The uniform corpus is 48 random bytes per key, so its trie is a wide shallow node whose chunks sit on or beside the single descent, and the cursor's read-ahead covers the rest. A seek can never cost more than the tree holds, so a 39-chunk tree caps it at 39.

Scale settles the law rather than one number. The same probe costs 7.0 fetches over 262 chunks at 1e4 and 21.8 fetches over 3910 chunks at 1e5, so the touched fraction collapses from 100% to 2.7% to 0.6% while N grows a hundredfold. A real whole-tree walk would hold that fraction at 1.0 and grow tenfold with the corpus. `the_ldb_seek_is_not_a_whole_tree_walk` changes N and fails any seek whose cost tracks it, and `the_ldb_ceiling_is_labelled_and_stays_seek_grade_across_scales` holds the ceiling to the same law while asserting it carries its read-ahead label.

## Gate

`cargo fmt` is clean for this crate; the remaining workspace drift is #644's and predates the branch. `cargo clippy -p nectar-manifest-sim --all-features --all-targets --no-deps -- -D warnings` is clean. `cargo nextest run -p nectar-manifest-sim` is 45 green in 42 seconds, including both determinism byte-identity gates. `cargo test --doc -p nectar-manifest-sim` passes. A round trip at `--scales 1000 --build-samples 1` writes the JSON and renders it back to markdown.

AI Assistance: Implement: claude-opus-5 (units A-F, rebase and honesty fixes), PR: claude-opus-5


mfw78:
Closing unmerged, to be re-cut as #680 against the landed seam.

Two reasons, both structural rather than a judgement on the benchmark itself. It depends on `nectar-loadsave`, which #660 dissolved into `mantaray::persist`, so its base no longer exists. And it carries 76 files of out-of-scope churn, more than the simulation crate it adds, which makes it unreviewable as a unit.

#680 re-cuts the two-arm mantaray 0.2 versus ldb harness onto the current tree, and folds it together with the `feeds-bench` and `ldb-sim` harness merge, which share a five-module layout and collide on `Corpus`, `Document` and `Meta`. The measurement intent is preserved there.

PR #726, opened 2026-08-12
ci: add cargo-deny job (advisories, bans, sources, licenses)

### LINKED ISSUE #700: ci: run cargo deny so deny.toml stops being dead config
Part of epic #697. Car 2 of the stage 1 serial stack: base this branch on the tip of #698, not on main, because both edit `.github/workflows/lint.yml` and the same `needs:` list.

## What
`deny.toml` is committed and `cargo-deny` is in the dev shell (`flake.nix:60`), but no workflow invokes it, so the file is dead config. It carries `[advisories]` and `[bans]` only; there is no `[licenses]` and no `[sources]` section.

Add a `deny` job to `.github/workflows/lint.yml` that installs `cargo-deny` through the pinned `taiki-e/install-action` pattern `audit.yml:31-34` already uses, and runs `cargo deny --all-features check advisories bans sources licenses`. The job needs no Rust toolchain: cargo-deny reads `Cargo.lock`.

Scope the overlap honestly. `audit.yml` already runs `cargo audit` over both lockfiles daily and on every manifest or lockfile change, and `deny.toml:4-5` records that its ignore list mirrors `.cargo/audit.toml`. The new coverage is `bans`, `sources`, `licenses`, and `yanked = "deny"` (`deny.toml:9`) on every pull request rather than only on lockfile-touching ones.

Keep the two ignore lists in step: an advisory added to `.cargo/audit.toml` goes into `deny.toml` as well, or the two gates disagree and one of them is noise.

## Measured on origin/main at 2731b110, 2026-08-12

`cargo deny --all-features check advisories bans sources` already passes: `advisories ok, bans ok, sources ok`. Every entry in `Cargo.lock` resolves to crates.io and there are no git dependencies, so `sources` passes on defaults.

That run emits one warning: `warning[advisory-not-detected]` at `deny.toml:19`, because the ignored advisory `RUSTSEC-2024-0388` no longer matches any crate in the resolved graph. The #659 fix bumped `ruint` to 1.20.0, which dropped the optional `ark-ff` path that pulled `derivative` in.

Do not remove the matching entry from `.cargo/audit.toml`. `derivative` is still present in both `Cargo.lock:1564` and `fuzz/Cargo.lock:1175`, and `cargo audit` scans the whole lockfile rather than the resolved graph, so it would still report the advisory and the `audit` job would go red. #703 is about to make that job blocking, so this matters.

Two acceptable outcomes; pick one and say which in the pull request body. Either remove the entry from `deny.toml` only and reword the header comment at `deny.toml:4-5`, which currently promises the two lists mirror each other, to record that deny scans the resolved graph while audit scans the lockfile and that the lists therefore diverge by design. Or leave both files untouched, because `advisory-not-detected` is a warning and `check advisories` still exits ok. Do not remove it from `.cargo/audit.toml` under either outcome.

`cargo deny --all-features check licenses` fails with 255 rejections. That number is not 255 problem dependencies: there is no `[licenses]` section at all, so cargo-deny 0.19 rejects everything including nectar's own twenty-four AGPL crates. Writing the section is the whole of the work.

The complete set of licence expressions in the `--all-features` graph, measured, is:

```
AGPL-3.0-or-later            24 crates (every nectar crate, keccak-batch, bmt-wasm-demo)
Apache-2.0                  202
Apache-2.0 WITH LLVM-exception 7
MIT                         225
MIT-0                         1  (dunce)
BSD-2-Clause                  2  (zerocopy, zerocopy-derive)
BSD-3-Clause                  5  (keccak-asm, sha3-asm, subtle, num_enum, num_enum_derive)
CC0-1.0                       1  (dunce)
LGPL-2.1-or-later             2  (r-efi)
Unicode-3.0                   1  (unicode-ident)
Unlicense                     1  (memchr)
Zlib                          1  (foldhash)
```

There is no unlicensed crate and no `NOASSERTION` in the graph.

Four of those are reachable only through an SPDX `OR` that a permissive allow already satisfies, so they do not need allowing: `LGPL-2.1-or-later` (r-efi is `MIT OR Apache-2.0 OR LGPL-2.1-or-later`), `Unlicense` (memchr is `Unlicense OR MIT`), `CC0-1.0` and `MIT-0` (dunce is `CC0-1.0 OR MIT-0 OR Apache-2.0`), and `BSD-2-Clause` (zerocopy is `BSD-2-Clause OR Apache-2.0 OR MIT`).

Two are load-bearing and must be allowed explicitly. `Unicode-3.0` is joined by an SPDX `AND`, not an `OR`: unicode-ident is `(MIT OR Apache-2.0) AND Unicode-3.0`. `Zlib` is foldhash's only licence.

Derive the final list from a local `cargo deny check licenses` run rather than pasting the block above, then state in a `deny.toml` comment why each non-obvious entry is there. Do not add a blanket `allow-osi-fsf-free`; the point of the section is that the list is reviewed.

The workspace is AGPL-3.0-or-later and publishes seven crates, so an unreviewed dependency licence is a release risk rather than a style question.

## Kills
Removes a committed configuration file that no process reads. Makes a yanked dependency a reported failure on main between nightly audit runs, and an unreviewed dependency licence a reported failure at all. Resolves the stale ignore entry that cargo-deny already warns about, without breaking the lockfile-scanning `audit` job that still needs it. #703 is the issue that makes either report block the merge.

## Gate
`lint / deny` green on main with `advisories bans sources licenses` all checked, and `audit` unchanged. `.cargo/audit.toml` still lists `RUSTSEC-2024-0388` and `audit` stays green. The job id is appended to the `needs` list of the `lint success` aggregator created by #698.

### LINKED ISSUE #729: ci: pin the deny job's toolchain instead of relying on the runner's rustup
Follow-up from #700 (PR #726). Independent; based on main once the stage 1 stack lands.

## What

The `deny` job installs `cargo-deny` and runs it with no toolchain step. cargo-deny shells out to `cargo metadata`, so cargo must be on `PATH`. It works because `ubuntu-latest` ships rustup, which resolves `rust-toolchain.toml` and installs 1.94 plus the `rustfmt` and `clippy` components the job never uses, on every run. The job also has no registry cache, so `cargo metadata` fetches the sparse index and every crate source uncached.

It is correct but slow and non-deterministic: a rustup or image change alters what the job installs.

Pin a bare toolchain instead, as the `msrv` and `format` jobs already do:

```yaml
- uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88
  with:
      toolchain: "1.94"
```

Not the `rust-setup` composite: that carries sccache and a registry cache keyed for the compiling jobs, and a non-compiling job that warms it with an empty `~/.cargo` degrades their cache hit rate.

## Also

`[licenses] allow` carries a bare `AGPL-3.0-or-later`, so a third-party AGPL dependency passes review-free alongside the workspace's own crates. Tightening it needs `[licenses] exceptions` enumerating the workspace members. Decide whether that is worth the churn; the licence is the same either way, so this is about review signal, not compliance.

## Gate

`lint / deny` green with a pinned toolchain, and the job no longer installs clippy or rustfmt. The compiling jobs' cache hit rate is unchanged.

mfw78:
Closes #700
Closes #729

## Summary

Adds a `deny` job to `.github/workflows/lint.yml` running `cargo deny --all-features --locked check advisories bans sources licenses`, installed through the same pinned `taiki-e/install-action@43aecc8d…  # v2.83.2` that `audit.yml` uses, on a pinned `dtolnay/rust-toolchain` at 1.94 rather than `rust-setup` (the job compiles nothing, so warming the shared registry cache with an empty `~/.cargo` would cost the compiling jobs). Its id is appended to the `lint success` aggregator's `needs` list; the list is now `[clippy, unused-deps, deny]`.

Writes the `[licenses]` section `deny.toml` never had. The allow list was derived from a local `cargo deny --all-features list` run in the dev shell, and every entry is load-bearing (dropping it fails `check licenses`): `AGPL-3.0-or-later` (the workspace's own crates), `MIT` and `Apache-2.0` (the bulk of the graph), `BSD-3-Clause` (sole licence of subtle, keccak-asm, sha3-asm), `Unicode-3.0` (mandatory because unicode-ident is `(MIT OR Apache-2.0) AND Unicode-3.0`), and `Zlib` (sole licence of foldhash).

Adds a `[sources]` section requiring every dependency resolve to crates.io (`unknown-registry = "deny"`, `unknown-git = "deny"`, `allow-registry` limited to the crates.io index), written out explicitly rather than relying on the (already-deny) defaults.

Reworks the `[advisories]` ignore list so it no longer mirrors `.cargo/audit.toml`. `RUSTSEC-2024-0388` (derivative, unmaintained) is dropped from `deny.toml` because cargo-deny walks the resolved feature graph and derivative is no longer reachable there since ruint 1.20.0 dropped the optional ark-ff path; keeping the entry would only earn an `advisory-not-detected` warning. `.cargo/audit.toml` keeps the entry, with an updated comment explaining why the two files now diverge: cargo audit scans the whole `Cargo.lock` (including `fuzz/Cargo.lock`), where derivative is still present as an unreachable lockfile entry. The `paste` and `proc-macro-error2` unmaintained ignores are unchanged.

## Folded in from #729

The `deny` job originally had no toolchain step at all, relying on the runner's preinstalled rustup resolving `rust-toolchain.toml` to give `cargo metadata` a cargo binary. That worked, but it reinstalled 1.94 plus the `rustfmt` and `clippy` components the job never uses on every run, and left the result dependent on the runner image. The job now pins the toolchain explicitly and sets `RUSTUP_TOOLCHAIN` on the `cargo deny` step. The environment variable is required, not belt-and-braces: `rust-toolchain.toml` declares `components = ["rustfmt", "clippy"]` and wins over an installed default at cargo time, so without the override the component install returns. This is the same override the `msrv` job uses for the same reason.

The second half of #729 is deliberately not done. It proposed replacing the bare `AGPL-3.0-or-later` allow with `[licenses] exceptions` enumerating the workspace members, so a third-party AGPL dependency could not pass unreviewed. That is real churn, it needs an edit every time a crate is added or removed, and it changes review signal rather than compliance, since the workspace licence is AGPL either way. Recorded here as a decision rather than an oversight; reopen if the trade-off is judged differently.

## Testing

`cargo deny --all-features --locked check advisories bans sources licenses` (the exact command the new `deny` job runs) exits 0.

`actionlint` is clean on `.github/workflows/lint.yml`, and the `deny` job's step list parses as checkout, toolchain, install-action, `cargo deny`, with `RUSTUP_TOOLCHAIN: "1.94"` on the final step.

## AI Assistance

Implementation: claude-opus-5. Red-team: claude-opus-5. PR: claude-sonnet-5.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -13,7 +13,11 @@ ignore = [
     # (via `ark-ff`) that is never enabled by the feature set alloy-primitives
     # requests, so it is not in the resolved build graph at all; it only
     # appears in Cargo.lock. cargo-audit scans the whole lockfile, which is
-    # why it surfaces here. Tracking: nxm-rs/nectar#19.
+    # why it surfaces here. deny.toml deliberately does not carry this entry:
+    # cargo deny walks the resolved graph, so the advisory never matches there
+    # and listing it would earn an `advisory-not-detected` warning. Removing it
+    # from this file would turn the `audit` job red, because derivative is still
+    # an entry in both Cargo.lock and fuzz/Cargo.lock. Tracking: nxm-rs/nectar#19.
     "RUSTSEC-2024-0388", # derivative unmaintained - optional ark-ff dep of ruint, not in build graph
 
     # paste is unmaintained (archived by its author). It is a compile-time
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -6,22 +6,21 @@
 # through the alloy stack. None of them are direct nectar dependencies, and
 # none are reachable at runtime or in the shipped wasm artifact. They are
 # kept here (rather than removed) because there is no released upstream
-# version that drops them yet. Each ignore links the upstream tracking issue
-# so it can be revisited once alloy publishes a release with the fix.
+# version that drops them yet. Revisit on each alloy bump.
 ignore = [
     # derivative is unmaintained. It is an *optional* dependency of `ruint`
     # (via `ark-ff`) that is never enabled by the feature set alloy-primitives
     # requests, so it is not in the resolved build graph at all; it only
     # appears in Cargo.lock. cargo-audit scans the whole lockfile, which is
     # why it surfaces here and why deny.toml omits it. Do not remove: it is
-    # still in both lockfiles. Tracking: nxm-rs/nectar#19.
+    # still in both lockfiles.
     "RUSTSEC-2024-0388", # derivative unmaintained - optional ark-ff dep of ruint, not in build graph
 
     # paste is unmaintained (archived by its author). It is a compile-time
     # proc-macro dependency of alloy-sol-macro-input / syn-solidity. The latest
     # released alloy-core (1.6.0) still depends on it; the drop to a maintained
     # alternative only exists on alloy-core's unreleased main branch. Build-time
-    # only, never shipped. Tracking: nxm-rs/nectar#20.
+    # only, never shipped.
     "RUSTSEC-2024-0436", # paste unmaintained - via alloy-sol-macro-input / syn-solidity
 
     # proc-macro-error2 is unmaintained but pulled in by alloy-sol-macro &
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -13,11 +13,8 @@ ignore = [
     # (via `ark-ff`) that is never enabled by the feature set alloy-primitives
     # requests, so it is not in the resolved build graph at all; it only
     # appears in Cargo.lock. cargo-audit scans the whole lockfile, which is
-    # why it surfaces here. deny.toml deliberately does not carry this entry:
-    # cargo deny walks the resolved graph, so the advisory never matches there
-    # and listing it would earn an `advisory-not-detected` warning. Removing it
-    # from this file would turn the `audit` job red, because derivative is still
-    # an entry in both Cargo.lock and fuzz/Cargo.lock. Tracking: nxm-rs/nectar#19.
+    # why it surfaces here and why deny.toml omits it. Do not remove: it is
+    # still in both lockfiles. Tracking: nxm-rs/nectar#19.
     "RUSTSEC-2024-0388", # derivative unmaintained - optional ark-ff dep of ruint, not in build graph
 
     # paste is unmaintained (archived by its author). It is a compile-time
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -123,14 +123,22 @@ jobs:
         timeout-minutes: 30
         steps:
             - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
-            # No `rust-setup`: this job compiles nothing.
+            # Not `rust-setup`: this job compiles nothing, so warming the shared
+            # registry cache with an empty ~/.cargo would cost the compiling jobs.
+            - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88  # master 2026-06-30
+              with:
+                  toolchain: "1.94"
             - uses: taiki-e/install-action@43aecc8d72668fbcfe75c31400bc4f890f1c5853  # v2.83.2
               with:
                   tool: cargo-deny
             # `--all-features` so the graph reaches feature-gated dependencies.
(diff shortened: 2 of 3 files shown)


mfw78:
Follow-up filed for the deferred findings from this PR: #729 (pin the deny job's toolchain; blanket AGPL allow).

PR #732, opened 2026-08-12
test: ruleset probe, markdown only (do not merge)

mfw78:
Temporary probe verifying that a change touching no code still produces all three required contexts, rather than stranding one at `Expected` forever.

None of `unit.yml`, `nostd.yml` or `lint.yml` carries a `paths:` filter, so all three should run and pass.

Expected result: all three required contexts report, and the pull request is not blocked. This pull request will be closed and its branch deleted once observed; it must never be merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,5 @@ AGPL-3.0-or-later. See [LICENSE](./LICENSE).
 ```
 ●  AGPL-3.0  ·  pre-release  ·  substrate under vertex
 ```
+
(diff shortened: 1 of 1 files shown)


mfw78:
Probe complete. All three required contexts (`unit success`, `no_std success`, `lint success`) reported and passed on a markdown-only change, so none of them can strand at `Expected`. The residual block is the pre-existing one-approval rule, not a status check. Closing unmerged.

Issue #716, opened 2026-08-11
release: derive the first-publish order from cargo metadata


mfw78:
Parent: #805, on the `M6 release 0.5.0` milestone. Everything else in this epic reads the order this issue writes, so it lands first inside the epic.

The two halves of this issue have very different shapes, and they can be split.
The publish-flag half is trivial, independent and can land at any time: it is two `publish = false` keys and no code.
The ordering half enumerates the tree, so it cannot be written until the crate set settles at the end of M2. Write it once, at the end, from `cargo metadata`.

## What

Cut the release surface, then write the order down once.

### Cut the set

`nectar-feeds` and `nectar-ldb` have no consumer outside this workspace, so set `publish = false` on both, exactly as `crates/testing/Cargo.toml:15` and `crates/integration-tests/Cargo.toml:11` already do. `nectar-envelope` parks the same way under decision 8 of `docs/RESTRUCTURE-PLAN.md`; M1 owns that one, not this issue. No crates.io name is then taken for code that nothing outside this workspace can use yet.

### Write the order down once, and only once

Merged pull request #497 (2026-07-25, `docs: drop the stale publish-order list from RELEASING.md`) deleted exactly this artefact from exactly this file. Its recorded reason: the list drifted three times in one week of normal work, it documented nothing cargo-release does not already compute, and the maintainer decision was to delete rather than guard. The restructure plan supersedes that decision for one release only, so the scope here is deliberately narrow: a one-shot 0.5.0 publish runbook that is deleted after the release, not a maintained section that re-enters the drift loop #497 killed.

Concretely, do not reinstate the deleted section and do not add a `First publish (0.5.0)` section in its place. Add `docs/FIRST-PUBLISH-0.5.0.md`, headed by a delete-on-completion notice that names this issue and the condition (every new name live on crates.io at 0.5.0), and link it from RELEASING.md with a single line that is removed in the same commit that deletes the file. #717 owns that deletion as its last step. RELEASING.md itself keeps #497's invariant: no crate count and no enumerated order.

Derive the order from `cargo metadata` over the tree as it stands when the runbook is written. Do not copy it from this issue, from #713, from #718 or from any other issue's prose. The derivation must cover every edge, including optional dependencies and versioned dev dependencies, because an optional edge still pins a version requirement that the registry must be able to satisfy.

[...]

Issue #713, opened 2026-08-11
docs: make the workspace crate table match the publish flags


mfw78:
Parent: #805, on the `M6 release 0.5.0` milestone. Independent; based on main.

## What

`README.md` lists nine crates in its table at lines 19 to 27. The workspace has 18 crates under `crates/*` plus the `crates/primitives/examples/wasm-demo` member, and 16 of the 18 carry no `publish = false`.

Four rows advertise crates.io and docs.rs badges for crates that have never been published, so both badges resolve to nothing: `nectar-primitives-core` at line 20, `nectar-manifest` at 22, `nectar-clock` at 26 and `nectar-tasks` at 27. Confirmed against the registry on 2026-08-21: all four return no crate.

Seven publishable crates have no row at all: `nectar-envelope`, `nectar-feeds`, `nectar-file`, `nectar-governor`, `nectar-ldb`, `nectar-marker` and `nectar-postage-usage`. Two of them are the largest crates in the workspace: `crates/ldb` at roughly 13,800 lines of Rust and `crates/file` at roughly 10,700. `nectar-postage-usage` is already published at 0.4.0 and still has no row.

Add the missing rows. Replace the badge cells on unpublished rows with an explicit "unpublished" marker until #716 first-publishes them. Then add a check that keeps the table honest: every workspace member without `publish = false` has exactly one row, and only rows for crates that are actually published carry badges. Run it in CI, so crate deletions and `publish = false` flips cannot silently desynchronize the table again.

The check must also cover the prose block under the table. `README.md:29-31` carries a deprecation note for `nectar-swarms`, which is no longer a workspace member and whose identifiers move again during the M2 carve. A note about a name that is not a member is exactly the kind of claim the check exists to catch, so decide whether the check owns it or the note is deleted.

The crate READMEs carry the same drift. All 16 publishable crates have a `README.md`. Twelve of them quote a version in the usage snippet, and five of those twelve are stale at `"0.1"` while the workspace version is `0.4.0` (`Cargo.toml:6`):

- `crates/contracts/README.md:11`
- `crates/mantaray/README.md:11`
- `crates/postage/README.md:11`
- `crates/postage-issuer/README.md:11`
- `crates/primitives/README.md:11`

[...]

PR #740, opened 2026-08-13
refactor: dissolve benches/examples crates into host crates

### LINKED ISSUE #618: phase 4: dissolve the benches and examples crates; codspeed to criterion

mfw78:
Closes #618

## Summary

Dissolved `crates/benches` and `crates/examples` into their host crates and swapped the codspeed shim for plain criterion.

Bench homing: `address`, `bmt_bench`, `encryption_bench`, `primitives`, `proofs` to `nectar-primitives`; `latency_bench` to `nectar-file`; `mantaray_bench` to `nectar-mantaray`; `verify` to `nectar-postage`; `sign` and `upload_pipeline` to `nectar-postage-issuer`. Example homing: `basic_usage` and `builder_patterns` to `nectar-primitives`; `dump_deployments` to `nectar-contracts`; `stamp_pipeline` to `nectar-postage-issuer`; `roam_between_machines` to `nectar-postage-usage`. All fifteen files moved with `git mv` so history follows.

Feature gating uses `required-features`, not a self dev-dependency. A self dev-dependency (the tokio pattern) would have unified the extra feature into every build of the host package that activates dev-dependencies, silently turning the default-feature CI matrix rows into feature-on rows.

`crates/benches` and `crates/examples` are removed as workspace members; their `Cargo.toml` and `src/lib.rs` are deleted.

## Testing

All commands run inside `nix develop --command` at `2c18d09c` (branch `refactor/618-dissolve-bench-crates`, based on `origin/main` @ `1ea486b3`). Toolchain: rustc 1.94.0, which equals the declared MSRV (`.github/scripts/declared-msrv.sh` -> `1.94`), so the build below doubles as the MSRV lane's `cargo check --workspace --all-targets --locked`.

**Workspace member count (`cargo metadata --no-deps`):** 23 before -> 21 after. `nectar-benches` and `nectar-examples` are the two removed members.

**Build and target relocation**
- `cargo build --workspace --all-targets --locked`: pass. Every moved bench and example builds in its new host crate.
- `cargo bench --workspace --locked --no-run`: pass (default-feature benches).
- `cargo bench --locked -p nectar-postage --features parallel --no-run`: pass (`verify`).
- `cargo bench --locked -p nectar-mantaray --features hazmat`: pass (`mantaray_bench`).
- All ten benches and five examples compile in their new hosts.
- All 14 clippy rows in `lint.yml` pass with `-D warnings`.
- `cargo fmt`, `cargo deny check` (advisories/bans/sources/licenses), and `cargo-machete` are clean.
- `nextest` is 573/573 on the seven touched crates.
- `cargo run -p nectar-contracts --example dump_deployments` emits the JSON `upstream-addresses.yml` consumes.
- `nostd.yml` uses plain `cargo check`; no bench target can leak onto the bare-metal lane.

## AI Assistance

Implemented by claude-opus-5. Red-teamed by claude-opus-5. PR description by claude-sonnet-5.


mfw78:
Correction to the member count, for the record, since stage 2's gate is that the final count is written down and correct.

The brief I gave said 22 members before and 20 after. That was wrong in both directions, and the red-team caught it. The workspace declares `members = ["crates/*", "crates/primitives/examples/wasm-demo"]`, so the member count is the number of directories under `crates/` plus one: `crates/primitives/examples/wasm-demo` is an explicitly listed member that a directory count misses.

The true figures are 23 members before this pull request and **21 after**. The directory count under `crates/` did go 22 to 20, which is where my number came from, but that is not the member count.

This matters beyond bookkeeping: stage 2's exit gate in `docs/PRODUCTION-PLAN.md` is that `cargo metadata` reports the final member count and it is written down, and #713 and #716 both enumerate the tree for the release. Anything quoting 20 should quote 21.

Issue #745, opened 2026-08-14
primitives: rename the ChunkRegistry envelope associated type and its accessors


mfw78:
Part of epic #802, in milestone M3. It was filed under #677, which is closed, and split out of #708 on 2026-08-14 because it is a workspace-wide sweep, not a primitives-local rename.

## What

`ChunkRegistry` declares an associated type named `Envelope` (`crates/primitives-core/src/chunk/registry.rs:90`), meaning "the closed union of chunk types this registry admits". `Chunk` exposes it through three accessors: `from_envelope` (`crates/primitives-core/src/chunk/trust.rs:217`), `envelope` (`:254`) and `into_envelope` (`:259`).

That word already means two other things. It is the HPKE sealed record for trojan chunks, which now lives in its own crate at `crates/envelope/`. In the reference client it is the HTTP route that mints a detached postage stamp, which `crates/postage/README.md:16-19` documents. Neither is a registry member union.

Rename the associated type and its accessors to something that says what they are.

## Why this is not a two-crate pass

Measured on `main` at 2026-08-21: `envelope` or `Envelope` appears on **137 lines across 46 files in 10 crates plus `fuzz/`**. The crates are `primitives-core`, `primitives`, `ldb`, `mantaray`, `file`, `feeds`, `postage`, `postage-issuer`, `testing` and `integration-tests`, plus `fuzz/fuzz_targets/ldb_build_roundtrip.rs:46`.

It also touches the compiled doctest at `crates/primitives-core/src/chunk/mod.rs:200`, which declares `type Envelope = ForkChunk;` inside the module walkthrough spanning `:72` to `:259`, and three README files: `crates/primitives-core/README.md:7`, `crates/primitives-core/src/chunk/README.md:15` and `:46`, and `crates/postage/README.md:16-19`.

**Correction: the HPKE envelope has moved.** The original body located it at `crates/primitives/src/envelope/`, which no longer exists. It is now the standalone `nectar-envelope` crate at `crates/envelope/`, which decision 8 of `docs/RESTRUCTURE-PLAN.md` parks at `publish = false`. Every remaining `envelope` reference inside `crates/primitives/src/` is the registry sense and is in scope for this rename; they are all `Chunk::from_envelope` calls in the store test modules.

That is why it is not scheduled beside other work. It previously collided with both other tracks, including `crates/ldb/src/store.rs` and `crates/postage/src/sink.rs`.

## The name needs deciding before any code

[...]

PR #749, opened 2026-08-14
refactor(envelope)!: extract the messaging crypto into nectar-envelope

### LINKED ISSUE #747: envelope: extract the messaging crypto into nectar-envelope and fix the inverted naming
Part of epic #677. Must land before 0.5.0 publishes. Depends on #708 only for merge order, not for content.

## What

Move the messaging crypto out of `nectar-primitives` into a new crate `nectar-envelope`, and rename its types so the vocabulary says what each thing is.

**Move**, together, because `envelope/compat.rs` is a 50-line adapter over `ecies`:

- `crates/primitives/src/envelope/` (1849 lines: `mod.rs` 941, `hpke.rs` 657, `attestation.rs` 201, `compat.rs` 50)
- `crates/primitives/src/ecies.rs` (405 lines)
- `EciesError` from `crates/primitives-core/src/error.rs:147`

**Rename**, because the current names are inverted. `Record` is the container and `Envelope` names the cipher suite, which is backwards: an envelope is a container, not a scheme.

| Today | Becomes | Why |
|---|---|---|
| `Record<'a>` | `Envelope<'a>` | the sealed container is what an envelope is |
| `SealedRecord<S>` | `SealedEnvelope<S>` | |
| `Opened<S>` | `Opened<S>` or `OpenedEnvelope<S>` | implementer's choice, state which |
| `SchemeId::Envelope` | `SchemeId::Hpke` | it discriminates the scheme, opposed to `Compat` |
| `EnvelopeOnly` | `HpkeOnly` | |
| `EnvelopeThenCompat` | `HpkeThenCompat` | |
| `EnvelopeAttestation` | `RecipientAttestation` | it attests a recipient's key, not an envelope |

`Topic`, `Recipient`, `Inbox`, `Compat` and `Hpke` keep their names. Note a type named `Hpke` already exists at `envelope/mod.rs:156`; reconcile it with `SchemeId::Hpke` and say how.

## Why this is misplaced

`nectar-primitives` is the layer-1 crate: chunks, addresses, BMT, the store traits. Sealing a message to a recipient under a topic is a layer-2 capability, exactly like `feeds`, `file` and `manifest`, all of which are peer crates that depend on primitives rather than living inside it.

The dependency evidence agrees. `ecies` has **exactly one consumer in the workspace: `envelope`**. Nothing else in any crate references it. And unlike `envelope` it is **ungated**, `pub mod ecies` at `crates/primitives/src/lib.rs:72`, so `nectar-primitives` unconditionally ships 405 lines of messaging crypto that only a feature-gated, zero-consumer module uses.

What the new crate needs from primitives is small: `chunk::encryption::EncryptionKey`, `error::WrongLength`, and the `Salt`/`Hint` vocabulary that moves with it. That is a normal layer-2 dependency, the same shape `feeds` has.

## Why before 0.5.0

Neither module exists at the `v0.4.0` tag, so moving them now costs nothing. After 0.5.0 publishes, `nectar_primitives::envelope::*` and `nectar_primitives::ecies::*` are public API and the move needs a major bump.

The same argument applies with more force to `PrimitivesError`. `EciesError` sits in `primitives-core` purely because `PrimitivesError` wraps it, which the type's own doc comment admits. Once `ecies` leaves, that variant must go or the graph becomes circular. `primitives-core` is not in the `v0.4.0` tag either, so removing the variant is free exactly once.

## Constraints

**no_std must survive.** `.github/workflows/nostd.yml:91-97` checks `-p nectar-primitives --no-default-features --features envelope` on `riscv64imac-unknown-none-elf`, because envelope open and decap need no RNG and are meant to work in a proving guest. That call site becomes `-p nectar-envelope --no-default-features`, and the new crate must be bare-clean on rv64. Sealing rides `encryption` (rand) and stays excluded, as today.

**The optional dependencies move with it.** `chacha20poly1305`, `hkdf` and `sha2` are envelope-only in `crates/primitives/Cargo.toml:37,38,40`; the first two leave `nectar-primitives` entirely, `sha2` stays because `nectar-ldb` uses it. The three `hpke-rs*` dev-dependencies move too: the four-way differential validation at `envelope/hpke.rs:430-535` is the crate's provenance story and must keep running.

**Do not weaken the tests.** 28 envelope tests pass today. They move; none is deleted or relaxed.

## Knock-ons to carry

- The workspace goes from 18 members to 19. `docs/PRODUCTION-PLAN.md` records 18 as stage 2's settled count; update it, keeping the note that the count is the `crates/` directory count plus one.
- #716 decides whether `nectar-envelope` publishes in 0.5.0 or is marked `publish = false` like `feeds` and `ldb`. Leave a comment there. It has no consumer, which is that issue's own test.
- `AGENTS.md:108` cites `crates/primitives/src/envelope/mod.rs:580` as the conforming example for the error-source rule. Repoint it.
- #710 commissions the normative specification and names the file `docs/spec/envelope-v1.md`. That filename survives this change; the crate name does not alter the format name or the frozen `nectar/env/v1` domain label.
- #708 marks the module unstable via an `envelope-unstable` feature on `nectar-primitives`. That feature disappears with the module. The other four parts of #708 (the IANA correction, the sender-authentication wording, the `Default*` dedupe, the postage README note) are unaffected and stay.
- The frozen wire is untouched: `nectar/env/v1`, the frame `hint || nonce || enc_x || ct`, and the compat byte-for-byte ECIES construction all stay exactly as they are.

## Gate

`cargo metadata` reports 19 members and the number is written into the plan.

`rg -n "ecies|envelope" crates/primitives/src crates/primitives-core/src` returns nothing outside a changelog. `PrimitivesError` has no `Ecies` variant. `nectar-primitives` no longer depends on `chacha20poly1305` or `hkdf`, proven by `cargo rustc --locked -p nectar-primitives --lib -- -D unused_crate_dependencies`.

The 28 envelope tests pass in the new crate, including the four hpke-rs differential directions and both `compile_fail` doctests. `cargo check --locked -p nectar-envelope --no-default-features --target riscv64imac-unknown-none-elf` is green, and `nostd.yml` names the new crate rather than the old feature.

No identifier named `Record` survives in the crate, and no `Envelope`-named identifier denotes a scheme rather than a container.

mfw78:
Closes #747

## Summary

Extracts the messaging crypto out of `nectar-primitives` into a new workspace member, `crates/envelope` (`nectar-envelope`).

Two commits, tip `440ca068` (`origin/main..HEAD`):

- `a8697739` `refactor(envelope)!: extract the messaging crypto into nectar-envelope`: the move itself.
- `440ca068` `build(primitives): keep the docs.rs feature list explicit`: follow-up doc/build tidy-up.

## The move

`crates/primitives/src/envelope/{mod,hpke,compat,attestation}.rs` and `crates/primitives/src/ecies.rs` become the new workspace member `crates/envelope` (`nectar-envelope`), with `envelope/mod.rs` landing as the crate root `src/lib.rs` and `ecies` staying a public module of it.

`EciesError` leaves `crates/primitives-core/src/error.rs` and is now declared in `crates/envelope/src/ecies.rs`; `PrimitivesError` loses its `Ecies` variant.

`chacha20poly1305`, `hkdf`, `sha2`, and the three `hpke-rs*` dev-dependencies move with the code, and the `envelope-unstable` feature disappears with the module rather than being preserved.

The three off-limits crates (`manifest`, `ldb`, `mantaray`) are untouched.

## Renames

`Record` -> `Envelope`, `RecordTooShort` -> `EnvelopeTooShort`, `SealedRecord` -> `SealedEnvelope`.

## Testing

All commands run inside `nix develop --command` at `440ca068` (`origin/main..HEAD` = 2 commits), worktree clean, nothing amended or pushed.

**Clippy (lint.yml parity)**
- `cargo clippy --workspace --all-targets --locked -- -D warnings`: clean
- `cargo clippy --locked --all-targets -p nectar-envelope --features encryption -- -D warnings`: clean (this is the pass `lint.yml` now runs in place of the old `-p nectar-primitives --features envelope-unstable,encryption`)
- `cargo clippy --locked --all-targets -p nectar-envelope -- -D warnings`: clean
- `--all-features` deliberately not run (`compile_error!` guards make it fail by construction on `main` too; CI does not use it)

**Unused dependencies**
- `cargo rustc --locked -p nectar-primitives --lib -- -D unused_crate_dependencies`: clean, confirming `chacha20poly1305`, `hkdf`, `sha2`, and the direct `k256` dep move with the code rather than lingering unused in `nectar-primitives`.

**Red-team verification**
- Re-derived the branch rather than trusting it. The move is faithful: every wire literal and pinned KAT frame is byte-identical to `origin/main`, and a LABEL mutation proves the KAT and the new `frozen_wire_constants` test actually bite.
- 39 tests (38 moved plus 1 new) pass in both the default and the `encryption` configuration; none was deleted or relaxed.
- The `compile_fail` doctest is genuinely non-vacuous under `--features encryption`, proven by inverting the recipient type.
- Gate items check out: 19 workspace members matching the plan's crate list, no `Record` identifier anywhere in the crate, no `Ecies` variant left in `PrimitivesError`.

## AI Assistance

Implementation: claude-opus-5. Red-team review: claude-opus-5. PR: claude-sonnet-5.

## Review follow-ups, added after the first pass

**One word now means two things, deliberately, and the issue's gate wording was wrong.** #747 asked that `rg -n "ecies|envelope"` over `crates/primitives/src` and `crates/primitives-core/src` return nothing. `ecies` does. `envelope` cannot: it is a pre-existing and unrelated word in `primitives-core`, naming the chunk registry's closed member union (`ChunkRegistry::Envelope`, `Chunk::from_envelope`, and about forty doc mentions). Nothing here touches it.

So after this merges the workspace carries `nectar_envelope::Envelope`, a sealed message container, beside `ChunkRegistry::Envelope`, a decoded chunk carrier. That is a genuine live collision rather than the awkward metaphor it was when #745 was filed, which makes #745 (rename the registry's associated type to `Member`) worth more than its original framing suggested.

**A factual correction to the module doc**, carried in the second commit. The doc said the ephemeral y-parity bit rides the low bit of `nonce[0]`, which is true of this crate but reads as a claim about the wire generally. The reference client mines it into `nonce[28]` and reads it back as `chunkData[36]` (`bee/pkg/pss/trojan.go:209-212`, `:259`).

Neither side breaks. Compat ECDH is x-only, so parity never reaches the shared secret, and bee's own reader forces odd regardless: `chunkData[36] | 0x1 != 0` is a bitwise or and is therefore always true. The divergence matters only to a future scheme that makes parity load-bearing on the compat frame, and it is exactly the kind of thing a test against reference-client bytes would pin. No such test exists yet; that gap is now recorded on #752.

### DIFF
--- a/crates/postage/README.md
+++ b/crates/postage/README.md
@@ -20,7 +20,6 @@ The reference client calls a detached postage stamp an envelope.
 The response carries the issuer address, the packed bucket and index, the timestamp and the signature.
 There is no Go type of that name; it is a route and a response name.
 `Stamp` in this crate is the stamp behind that response, field for field: batch, index, timestamp and signature.
-The `envelope` module in `nectar-primitives` is unrelated: it is the HPKE encryption envelope.
 
 ## License
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -86,10 +86,10 @@ jobs:
             # likewise outside the default-features pass.
             - name: cargo clippy (ldb encryption)
               run: cargo clippy --locked --all-targets -p nectar-ldb --features encryption -- -D warnings
-            # The modern envelope sits behind the primitives envelope-unstable
-            # feature, likewise outside the default-features pass.
+            # The envelope sealing lane sits behind the encryption feature,
+            # likewise outside the default-features pass.
             - name: cargo clippy (envelope)
-              run: cargo clippy --locked --all-targets -p nectar-primitives --features envelope-unstable,encryption -- -D warnings
+              run: cargo clippy --locked --all-targets -p nectar-envelope --features encryption -- -D warnings
             # The pipeline's inline engine only compiles with the parallel
             # feature off, so lint the inline shape on its own.
             - name: cargo clippy (postage-issuer inline engine)
@@ -103,7 +103,7 @@ jobs:
             - name: unused lib dependencies
               run: |
                   for crate in nectar-primitives nectar-primitives-core \
-                               nectar-postage \
+                               nectar-envelope nectar-postage \
                                nectar-postage-issuer nectar-postage-usage \
                                nectar-contracts nectar-file \
                                nectar-marker nectar-clock nectar-governor \
--- a/.github/workflows/nostd.yml
+++ b/.github/workflows/nostd.yml
@@ -93,8 +93,8 @@ jobs:
             - name: Check envelope open builds bare-metal
               if: matrix.target == 'riscv64imac-unknown-none-elf'
               run: |
-                  cargo check --locked -p nectar-primitives \
-                    --no-default-features --features envelope-unstable \
+                  cargo check --locked -p nectar-envelope \
+                    --no-default-features \
                     --target ${{ matrix.target }}
             # The single-thread send escape (the unsync feature) and its
             # !Send-store compile proof, checked for the host inside the
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -119,16 +119,15 @@ jobs:
                   cargo nextest run \
                     -p nectar-mantaray --features manifest,encryption --locked \
                     --no-tests=warn --no-fail-fast
-            # The modern envelope (KAT, differential and forgery batteries)
-            # sits behind the primitives envelope-unstable feature; cover its
-            # tests and the no-downgrade compile-fail doctest here.
+            # The envelope sealing lane sits behind the encryption feature;
+            # cover it and the no-downgrade compile-fail doctest here.
             - name: Run envelope tests
               run: |
                   cargo nextest run \
-                    -p nectar-primitives --features envelope-unstable,encryption --locked \
+                    -p nectar-envelope --features encryption --locked \
                     --no-tests=warn --no-fail-fast
             - name: Run envelope doctests
-              run: cargo test --doc -p nectar-primitives --features envelope-unstable,encryption --locked
+              run: cargo test --doc -p nectar-envelope --features encryption --locked
             # The encrypted allocation probes must see encryption without
             # rayon, matching the encrypted split shape above.
             - name: Run encrypted split integration tests
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -105,9 +105,9 @@ The error layer is where duplication now lives, so these six rules are the stand
 Each rule below names a file and line that already conforms, so a reviewer can check the claim rather than take it.
 
 - Rule 1: `crates/feeds/src/error.rs:13`, `#[non_exhaustive]` on `FeedError`.
-- Rule 2: `crates/primitives/src/envelope/mod.rs:580`, a `#[from]` wrapping variant.
+- Rule 2: `crates/envelope/src/lib.rs:598`, a `#[from]` wrapping variant.
 - Rule 3: `crates/feeds/src/error.rs:17`, `AddressMismatch` carrying typed `expected` and `actual` fields rather than a formatted string.
-- Rule 4: `crates/primitives-core/src/error.rs:163`, the `BoxedError` alias.
+- Rule 4: `crates/primitives-core/src/error.rs:142`, the `BoxedError` alias.
 - Rule 5: no conforming example exists yet.
   `RingExhausted` is declared in more than one crate with the same shape and no conversion between them, which is the violation this rule exists to stop; #688 gives it one owner.
 - Rule 6: `crates/postage-usage/src/error.rs:271` and `:330`, `UsageError::is_corruption` and `is_recoverable`, with the exhaustive classification test at `:387` that fails to compile when a variant is added without being classified.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2515,6 +2515,27 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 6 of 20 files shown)


crates/postage/README.md
```diff
@@ -20,7 +20,7 @@ The reference client calls a detached postage stamp an envelope.
 The response carries the issuer address, the packed bucket and index, the timestamp and the signature.
 There is no Go type of that name; it is a route and a response name.
 `Stamp` in this crate is the stamp behind that response, field for field: batch, index, timestamp and signature.
-The `envelope` module in `nectar-primitives` is unrelated: it is the HPKE encryption envelope.
+The `nectar-envelope` crate is unrelated: it is the HPKE encryption envelope.
```
mfw78: I think this line can probably just be deleted.

mfw78: Deleted.

The line came in with #746 and #749 was only repointing it at the new crate, so it now goes entirely rather than being reworded. The section around it exists to tell a reader of the bee API that the `/envelope` route response is `postage::Stamp` field for field, and it says that completely without the third cross reference. The line would also have aged badly: once the scheme seam lands, `nectar-envelope` holds more than HPKE, so "it is the HPKE encryption envelope" stops being true.

Head is now `6349cbc`, with #755 rippled on top.

Issue #770, opened 2026-08-14
postage-issuer: dedicated sign stage so put slots are pure put slots


mfw78:
Part of #760, wave 2. Depends on A3 and B3.

## What

A dedicated sign stage, expressed as a `Stream` adapter: chunks in, sealed pairs out, unordered completion, with its own `SignWindow` independent of the put window.

The put stage drains the sealed stream into any sink under `PutWindow`, so a put slot is occupied only for store latency. Backpressure propagates by chained admission: the put window fills, the sign stage's completion queue fills, `poll_admit` parks the admitter, and the splitter stalls. The buffer between stages is at most one sign window of stamps, which is about 29 KiB at 256, not chunks.

Reuse the existing machinery rather than duplicating it: the admission window, `FuturesUnordered`, the governor, the delivery guard that hands a signed stamp back on a failed or cancelled delivery, and the dropped-job mapping that stops a lost task wedging the window.

## Why

The split's put window defaults to 16 and those slots double as sign-plus-put slots against a validated sign window of 256, so a 50 ms signer caps throughput at roughly 320 chunks per second where 5120 is available. Separately, the current decorator calls `prepare_stamp` inside its own shared-state lock, so a concurrent issuer placed inside it gains nothing.

This is the contingency #487 named and #540 was meant to measure. Fixing it structurally retires the current guidance that callers should widen the put window by hand.

## Gate

A slow signer no longer moves put throughput. The existing idempotence and multiset contracts hold: one result per input, unordered, fail-fast with a not-admitted tail. #540 then measures the fix rather than the hazard.

mfw78:
Correction to this issue text, found by the red-team on pull request #785.

The claim that "the buffer between stages is at most one sign window of stamps, which is about 29 KiB at 256, not chunks" is wrong, and it was my error when writing this issue. The stage retains up to one sign window of chunks in flight plus one window of sealed pairs, so the real bound is on the order of 2 MiB of chunk handles at a window of 256, not 29 KiB.

The implementation states the true bound in its rustdoc, so only this issue text was wrong. Recording it here so the figure is not carried into #540 or into any later memory-bound work.

mfw78:
Rebased onto main as pull request #794, which supersedes #785. #785 was based on the stale `integration/tracks`; #794 is a single commit on current main carrying the owner-bound signer seam from #789.

One finding worth recording, raised by the red team and confirmed against the branch.

[...]

PR #776, opened 2026-08-14
refactor(postage)!: put trust and validation state on StampedChunk

### LINKED ISSUE #768: postage: StampedChunk over trust and validation state
Part of #760, track B. Depends on B1. Breaking for vertex, record on #719.

## What

Both facts on the pair.

```rust
pub struct StampedChunk<
    T: TrustState = Verified,
    V: ValidationState = Validated,
    const BODY_SIZE: usize = DEFAULT_BODY_SIZE,
> { chunk: Chunk<T, AnyChunkSet<BODY_SIZE>>, stamp: Stamp, _validation: PhantomData<V> }
```

Transitions: `parse` produces the doubly-unblessed pair; `new` exists only at `Unvalidated`, so a validated pair cannot be fabricated; `validate` and `verify` are consuming and advance one axis each; `detach` projects into `StampedAddress<V>` so there is one implementation of the check.

The two orders both work and both are meaningful. Ingest runs validate before verify, which is the anti-denial-of-service ordering `crates/postage/src/sink.rs` currently mandates in a doc comment and which this makes type-enforced. The producer path is already verified and only needs validation.

**Freeze the parameter order in this issue.** States first, matching `Chunk<S: TrustState, R>`. Twelve explicit turbofish sites change, most of them `impl` blocks inside `stamped.rs` that this issue rewrites anyway; thirty-six bare `StampedChunk` sites keep compiling because the defaults are the safe states. This is unfixable after publication.

## Why

Today the pair hard-pins `Chunk<Verified>`, so wire decode must verify eagerly and there is no way for a sink to demand that a stamp was checked. Validation is currently ephemeral: the validator returns `Result<(), StampError>` and discards the batch it just fetched.

## Gate

The nonsense edge is unrepresentable: a validated pair cannot be constructed except through a transition that used the pair's own address. The 113-byte encoding is unchanged, asserted by round trip. Fuzz corpora replay unchanged.

mfw78:
Closes #768

## What

Puts trust and validation state on `StampedChunk` as typestate generics, and adds a state-generic address accessor to `Chunk`.

- `crates/primitives-core/src/chunk/trust.rs`: `Chunk<S, R>` gains `address_in_state()`, a state-generic accessor that reads the address regardless of trust state. `claimed_address` and `address` stay pinned to their single states, so nothing existing is loosened.
- `crates/postage/src/stamped.rs`: `StampedChunk` becomes `StampedChunk<T: TrustState = Verified, V: ValidationState = Validated, const BODY_SIZE: usize = DEFAULT_BODY_SIZE>`, holding the chunk, the unchanged `Stamp`, and a `PhantomData<V>`. The parameter order is states first, matching `Chunk<S: TrustState, R>`, with defaults set to the safe states (`Verified`, `Validated`). `new` is only available at `Unvalidated`, so a validated pair cannot be fabricated directly; `validate` is the sole route from `Unvalidated` to `Validated` and certifies the stamp against a `Batch` and the chunk's own address.
- `crates/postage/src/sink.rs`: the sink traits are updated for the new generic shape.
- `crates/postage/src/generators.rs`: adds `batch_and_stamp`, factored out of `signed_stamped_chunk`, and updates the property test to also assert `stamped.validate(&batch).is_ok()`.
- `crates/postage/src/stamped_address.rs`, `crates/postage/src/lib.rs`: supporting type/doc updates for the new state parameters.
- `crates/postage-issuer/src/pipeline/stamped_put.rs`: `PutStamped` is now parametrised by validation state (`PutStamped<Unvalidated>`), since the decorator mints the stamp rather than checking it, so the pair it forwards carries no validation proof. Test sinks updated accordingly.

## Why

A state-generic carrier needs the chunk's address before the trust axis is settled, but `claimed_address`/`address` are each pinned to one state. `address_in_state()` fills that gap without loosening either existing accessor.

Putting trust and validation state directly on `StampedChunk` makes both axes visible in the type, so a validated, verified pair is unrepresentable except through the one function (`validate`) that actually checks it. This closes the previous gap where a `StampedChunk` carried a stamp with no compile-time record of whether it had been validated.

## Testing

- `cargo fmt --all --check`
- Correctness of the typestate was verified by exhausting the routes to `StampedChunk<_, Validated, _>`: `validate` is the only mint and reads the pair's own address; `verify` preserves `V`; `Chunk::verify` never rewrites `address`; `Clone` is state-preserving.
- New unit test `address_in_state_reads_both_states` in `crates/primitives-core/src/chunk/trust.rs`.
- `crates/postage/src/generators.rs` property test extended to assert `stamped.validate(&batch).is_ok()`.

AI Assistance: implement claude-opus-5, red-team claude-opus-5, PR claude-sonnet-5.

### DIFF
--- a/crates/postage-issuer/src/pipeline/stamped_put.rs
+++ b/crates/postage-issuer/src/pipeline/stamped_put.rs
@@ -234,8 +234,8 @@ enum Step {
 /// Stamping decorator over a [`PutStamped`] sink.
 ///
 /// Implements `ChunkPut` facing up and requires `P: PutStamped<Unvalidated>`
-/// facing down: the stamp is minted here rather than checked, so the pair
-/// carries no validation proof. One generic impl over `AnyChunkSet<B>` serves every
+/// facing down: the stamp is minted here, not checked, so the pair carries no
+/// validation proof. One generic impl over `AnyChunkSet<B>` serves every
 /// wrapped call site.
 ///
 /// # Contracts
@@ -256,8 +256,8 @@ enum Step {
 /// - Wrapping a purely local store burns indices for chunks that may never
 ///   reach the network; filter for presence upstream where that matters.
 /// - Put-only sites need `P: PutStamped<Unvalidated>`; commit and apply sites
-///   need `P: PutStamped<Unvalidated> + TrustedGet + ChunkHas`, so a pure network sender
-///   takes a local or teed inner there.
+///   need `P: PutStamped<Unvalidated> + TrustedGet + ChunkHas`, so a pure
+///   network sender takes a local or teed inner there.
 /// - A split's put slots double as sign-plus-put slots: widen the put
 ///   window toward [`StampPipeline`](super::StampPipeline)'s default
 ///   window when wrapping a slow signer, and prefer an owned clone
--- a/crates/postage-issuer/src/pipeline/stamped_put.rs
+++ b/crates/postage-issuer/src/pipeline/stamped_put.rs
@@ -15,7 +15,7 @@ use nectar_clock::Clock;
 #[cfg(feature = "std")]
 use nectar_clock::SystemClock;
 use nectar_marker::{MaybeSend, MaybeSync};
-use nectar_postage::{PutStamped, Stamp, StampDigest, StampError, StampedChunk};
+use nectar_postage::{PutStamped, Stamp, StampDigest, StampError, StampedChunk, Unvalidated};
 use nectar_primitives::{AnyChunkSet, Chunk, ChunkAddress, ChunkGet, ChunkHas, ChunkPut, Verified};
 
 #[cfg(feature = "std")]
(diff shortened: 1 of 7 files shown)


mfw78:
Integration finding, not a defect in this pull request: #781.

Track A and track B disagree about how a producer mints a validated pair. `Prepared::seal` expects to return the sealed pair, `StampedChunk::new` exists only at `Unvalidated` by design, and `Prepared` holds a `BatchId` rather than a `Batch<S>` so it cannot call `validate`. Each side is defensible alone; together there is no path from issuance to a validated pair, and the sink then demands one.

`integration/tracks` at `c65cd5c4` carries both tracks merged and gated green, with `seal` returning the unvalidated pair as the least committal resolution. That resolution is on the integration branch only and is not proposed here. Wave 2 (#770) is held until #781 is decided.

Issue #764, opened 2026-08-14
postage-issuer: lock-free per-bucket counters and atomic depth


mfw78:
Part of #760, track A. Depends on A1.

## What

Replace the sharded counter with lock-free allocation.

Per-bucket counters become a flat `[AtomicU32]` with a compare-and-swap allocation loop: load the count, load the depth, refuse if the count has reached capacity, otherwise compare-and-swap from `n` to `n + 1`. Compare-and-swap rather than fetch-add-then-roll-back, because rollback has a lost-slot race when two overshooters interleave with a dilution.

Depth becomes an `AtomicU8` advanced by `fetch_max`, which makes the non-decrease invariant lock-free and makes a redelivered or reordered chain event a no-op by construction.

Delete `crates/postage-issuer/src/sharded.rs` rather than extending it, per the delete-before-refactor rule.

The mutable-batch ring stays lock-based. Its protected-slot scan is multi-word and does not compare-and-swap, and mutable-batch mass signing is not the hot path.

Under `not(multi_thread)`, which covers `unsync`, wasm32 and bare metal, the arm stays the existing sequential `CounterTableFor` behind the same interface, so the single-threaded path pays no atomic cost. `portable-atomic` enters as the fallback for bare-metal targets without native compare-and-swap.

## Why

Allocation must not go through a mutex. The current design also allocates a full-width counter table per shard, about 4 MiB for mainnet, of which fifteen sixteenths is dead; a flat table is 256 KiB.

Safety rests on monotonicity: capacity only grows and counters never move, so a stale depth read yields a smaller bound and can only refuse conservatively, never over-issue.

## Gate

A property test that concurrent allocation hands out each slot exactly once. A test that a dilution observed mid-allocation never produces a double-spend. Snapshot persistence is a monotone under-approximation, which is accepted, so the SBU1 bytes and the restore path are unchanged. The riscv64 and wasm32 `no_std` lanes stay green.

PR #779, opened 2026-08-14
refactor(postage-issuer): permit seam for stamp reservation

### LINKED ISSUE #765: postage-issuer: the permit seam, reserve(&self) returning Prepared<S>
Part of #760, track A. Depends on A2. Breaking for vertex, record on #719.

## What

The permit seam. `StampIssuer::reserve(&self) -> Result<Prepared<S>, StampError>` replaces `prepare_stamp(&mut self)`.

`Prepared<S>` is the permit: constructing it consumed a slot. It carries the address, the batch, the `Bucket<S>` and the depth it was allocated under, and it owns its admission-window token.

Separate the three phases so signing never holds a lock: reserve is cheap and atomic, signing is expensive and happens outside every lock, and commit is bookkeeping.

`Prepared::seal(chunk, sig)` mints a sealed pair, checking that the chunk is the one the slot was allocated for.

Cancellation safety is by `Drop`, never by code after an await. Dropping a permit releases the admission-window token so backpressure recovers, and burns the slot, which is the existing documented contract and is what keeps the SBU1 monotone watermark expressible.

## Why

The `&mut self` borrow is what serializes allocation at the sink today, and it is also why an issuer cannot be diluted while a pipeline holds it. Moving to `&self` with interior mutability from A2 is the precondition for A4.

## Gate

A test that a permit dropped mid-flight releases its window token and leaves issuer state coherent. The burn-on-failure contract stated in the rustdoc of the permit type. The riscv64 and wasm32 `no_std` lanes stay green.

mfw78:
Closes #765

## What

Track A car 3 (#765), the permit seam. `StampIssuer::prepare_stamp(&mut self)` is gone; the trait now has an associated `Spec` and `reserve(&self, address, timestamp) -> Result<Prepared<Self::Spec>, StampError>`.

New module `crates/postage-issuer/src/permit.rs` carries three public types. `Prepared<S>` is the permit: constructing it consumed the slot, and it carries the address, the `BatchId`, the `Bucket<S>`, the `BatchDepth<S>` it was allocated under, the slot, the timestamp and an optional `WindowToken`. `AdmissionWindow` is a clone-shared counting window (atomic under `multi_thread`, `Cell` otherwise, `Arc`/`Rc` accordingly) whose `try_acquire` hands out `WindowToken`. `WindowToken::drop` returns the slot, so dropping a permit recovers backpressure and burns the batch slot; that contract is stated in the `Prepared` rustdoc.

`Prepared::seal(chunk, sig)` mints a `StampedChunk`. `crates/postage-issuer/src/prepared.rs` is deleted; its contents are superseded by `permit.rs`. `issuer.rs`, `ring.rs`, `pipeline/mod.rs`, `pipeline/stamp_sink.rs`, `stamper.rs`, `lib.rs`, `postage-usage/src/issuer.rs` and `postage/src/error.rs` are all updated to the new `reserve`/`Prepared` seam, along with the benches, examples and tests under `postage-issuer/`.

## Why

Moves the reservation of a stamp slot from a `&mut self` call into an immutable, permit-based API (`reserve` returns an owned `Prepared<Spec>`), so slot admission and release compose with the lock-free counters from #764 instead of requiring exclusive access to the issuer.

## Testing

All commands ran inside `nix develop --command` on the dev shell toolchain (rustc 1.94.0, which is exactly the declared MSRV, so the MSRV job's `cargo check --workspace --all-targets --locked` is covered by the workspace clippy pass below).

Clippy (lint.yml parity, no `--all-features` anywhere):
- `cargo clippy --workspace --all-targets --locked -- -D warnings` : clean.
- `cargo clippy --locked --all-targets -p nectar-postage-issuer --features parallel -- -D warnings` : clean.
- `cargo clippy --locked --all-targets -p nectar-postage-issuer -- -D warnings` (inline engine) : clean.
- `cargo clippy --locked --all-targets -p nectar-postage --features serde -- -D warnings` : clean.
- `cargo clippy --locked --all-targets -p nectar-postage --features parallel -- -D warnings` : clean.

## Red-team

Red-teamed refactor/765-permit-seam against origin/refactor/764-lockfree-counters. Re-derived the token accounting from the code rather than trusting the report: the claim that outstanding `WindowToken`s always equal `in_flight.len()` holds, because every token is captured by exactly one future's `map` closure and futures-util drops that closure at the same poll that removes the task from the set, so the `Poll::Ready(None)` arm of `poll_admit` is genuinely unreachable, not merely untested. Non-vacuousness of the new tests was checked by mutation, for example swapping `drop(token)` for a leak, to confirm they fail without the fix.

AI Assistance: implement claude-opus-5, red-team claude-opus-5, PR claude-sonnet-5.


mfw78:
Integration finding, not a defect in this pull request: #781.

Track A and track B disagree about how a producer mints a validated pair. `Prepared::seal` expects to return the sealed pair, `StampedChunk::new` exists only at `Unvalidated` by design, and `Prepared` holds a `BatchId` rather than a `Batch<S>` so it cannot call `validate`. Each side is defensible alone; together there is no path from issuance to a validated pair, and the sink then demands one.

`integration/tracks` at `c65cd5c4` carries both tracks merged and gated green, with `seal` returning the unvalidated pair as the least committal resolution. That resolution is on the integration branch only and is not proposed here. Wave 2 (#770) is held until #781 is decided.

PR #785, opened 2026-08-17
feat(postage-issuer): sign in a stage of its own so put slots are pure put slots

### LINKED ISSUE #770: postage-issuer: dedicated sign stage so put slots are pure put slots

mfw78:
Closes #770

## Summary

Two new surfaces in `nectar-postage-issuer`, both built on the existing machinery rather than a new substrate.

`SignStage` (`crates/postage-issuer/src/pipeline/sign_stage.rs`) is the dedicated sign stage: chunks in through `poll_admit`, sealed `StampedChunk` pairs out through `poll_next`, unordered, under the pipeline's own sign window. It wraps `StampSink`, so admission, `FuturesUnordered`, the governor's admission window, the permit's window token, and the dropped-job mapping are reused verbatim; the stage adds only the address-keyed queue of chunks awaiting their stamp and a sealed-pair buffer capped at one sign window. It also implements `Stream`, ending only after `close()` plus a drain, so it is not a stream that yields `None` and then more items.

`StagedPut` (`crates/postage-issuer/src/pipeline/staged_put.rs`) is the put stage: a `ChunkPut<Chunk<Verified, AnyChunkSet<B>>>` decorator that drains the sign stage into any `ChunkPut<StampedChunk>` inner sink, so put slots only ever see stamped work.

Supporting changes: `crates/postage-issuer/src/pipeline/shared.rs` (+73) factors out logic shared between the two stages, `pipeline/mod.rs` and `stamped_put.rs` are wired up accordingly, and `lib.rs` re-exports the new types.

## Fix applied during review

A real deadlock was found and fixed (commit `e687f151`): `StagedPut` relies on `FuturesUnordered` and `PutSink` retaining only their latest waker registration, with the settling poll handing the wake to the other parked pollers. A put or flush future dropped while parked broke that invariant: it could hold the sole live registration, and the next sign or delivery completion would wake a waker that no longer existed, stalling the pipeline. The fix hands the parked wake on explicitly when a staged put leaves.

A mechanical fix was also applied and pushed (commit `10c54c5a`): `cargo fmt --all --check` was red on the branch as pushed: five files in the touched crate were unformatted (`lib.rs` re-export ordering, `shared.rs`, `sign_stage.rs`, `staged_put.rs`, `stamped_put.rs`), which CI's `lint.yml` checks at line 214. Ran `cargo fmt --all` (14 insertions/14 deletions, whitespace and import-order only, no semantic change) and re-ran the verification battery afterwards.

Both fix commits are unsigned, matching the two existing branch commits (also unsigned).

## Testing

Full verification battery passed after both fixes were applied.

## AI Assistance

Implementation: claude-opus-5. Red-team review: claude-opus-5. PR: claude-sonnet-5.


mfw78:
Superseded by #794, which is the same work rebased onto current main. This branch was based on the stale integration/tracks.

